PR #51, opened 2023-06-19
✨ add conditions in status

pdettori:
## Related issue(s)

Fixes #


waltforme:
Quick question: is `api/v1alph1/conditions.go` written by some code-gen or created by hand?

pdettori:
It’s written by hand. Not sure there are generators to produce conditions. Some frameworks like Crossplane provide conditions when extending their API.

waltforme:
I see. Thanks for letting me know these!

pdettori:
You are welcome @waltforme

PR #175, opened 2024-01-21
✨ add chattyStatus flag

effi-ofer:
## Summary
Added a chattyStatus flag to kflex.  Once this is merged, we'll update the kubestellar CICD to use the flag so we won't get spammed in the testing output.

## Related issue(s)

Fixes #


pdettori (approved):
I have tested the CLI build from the PR branch and it works as expected. Thanks @effi-ofer !
/lgtm

PR #180, opened 2024-01-24
✨ Add PostgreSQL DB as a sub-chart of `kubeflex-operator`

francostellari:
## Summary

- PostgreSQL DB was added as a sub-chart of `kubeflex-operator` charts
- Helm packaging was correspondingly updated in the git actions
- Git actions were changed to include the correct `--appversion`
- Go code for `kflex init` was updated to match the new chart
- `user.md` documentation was updated to match the new chart

## Related issue(s)

Fixes #


cmd/kflex/init/init.go
```diff
@@ -50,16 +50,6 @@ func Init(ctx context.Context, kubeconfig, version, buildDate string, domain, ex
 		util.PrintStatus("OpenShift cluster detected", done, &wg)
 	}
 	done <- true
-	util.PrintStatus("Installing shared backend DB...", done, &wg)
-	ensureSystemDB(ctx, isOCP)
-	done <- true
-
-	util.PrintStatus("Waiting for shared backend DB to become ready...", done, &wg)
```
pdettori, line 57: Since Postgres is only used by control planes of type k8s (by default with shared DB), just the fact that the kubeflex controller manager is up does not guarantee that the postgres DB is up. I think we should still have this check after the chart is installed, as well as the check for the controller.

cmd/kflex/init/init.go
```diff
@@ -50,16 +50,6 @@ func Init(ctx context.Context, kubeconfig, version, buildDate string, domain, ex
...
-	util.WaitForStatefulSetReady(
-		*(client.GetClientSet(kubeconfig)),
-		util.GeneratePSReplicaSetName(util.DBReleaseName),
```
pdettori, line 60: Note that DBReleaseName needs to change from "postgres" to  "kubeflex-operator".

cmd/kflex/init/init.go
```diff
@@ -81,53 +71,27 @@ func Init(ctx context.Context, kubeconfig, version, buildDate string, domain, ex
...
+		vars = append(vars,
+			"postgresql.primary.podSecurityContext.fsGroup=null",
+			"postgresql.primary.podSecurityContext.seccompProfile.type=RuntimeDefault",
+			"postgresql.primary.containerSecurityContext.runAsUser=null",
+			"postgresql.primary.containerSecurityContext.allowPrivilegeEscalation=false",
+			"postgresql.primary.containerSecurityContext.runAsNonRoot=true",
+			"postgresql.primary.containerSecurityContext.seccompProfile.type=RuntimeDefault",
+			"postgresql.volumePermissions.enabled=false",
+			"postgresql.shmVolume.enabled=false",
+		)
```
pdettori, line 94: I am not completely sure that the parameters for OCP are passed correctly with the new setup. I did a quick test building the chart and pushing to a temp personal repo on quay.io, building the CLI from the PR branch and running `kflex init` vs the OCP cluster, and the postgres DB does not start - stateful set reports all these issues:

```
Warning  FailedCreate      35s (x16 over 118s)  statefulset-controller  create Pod kubeflex-operator-postgresql-0 in StatefulSet kubeflex-operator-postgresql failed error: pods "kubeflex-operator-postgresql-0" is forbidden: unable to validate against any security context constraint: [provider "anyuid": Forbidden: not usable by user or serviceaccount, provider "pipelines-scc": Forbidden: not usable by user or serviceaccount, provider restricted-v2: .spec.securityContext.fsGroup: Invalid value: []int64{1001}: 1001 is not an allowed group, provider restricted-v2: .containers[0].runAsUser: Invalid value: 1001: must be in the ranges: [1000730000, 1000739999], provider "restricted": Forbidden: not usable by user or serviceaccount, provider "ibm-restricted-scc": Forbidden: not usable by user or serviceaccount, provider "nonroot-v2": Forbidden: not usable by user or serviceaccount, provider "nonroot": Forbidden: not usable by user or serviceaccount, provider "ibm-anyuid-scc": Forbidden: not usable by user or serviceaccount, provider "hostmount-anyuid": Forbidden: not usable by user or serviceaccount, provider "log-collector-scc": Forbidden: not usable by user or serviceaccount, provider "ibm-anyuid-hostpath-scc": Forbidden: not usable by user or serviceaccount, provider "hostnetwork-v2": Forbidden: not usable by user or serviceaccount, provider "hostnetwork": Forbidden: not usable by user or serviceaccount, provider "hostaccess": Forbidden: not usable by user or serviceaccount, provider "ibm-anyuid-hostaccess-scc": Forbidden: not usable by user or serviceaccount, provider "node-exporter": Forbidden: not usable by user or serviceaccount, provider "ibm-privileged-scc": Forbidden: not usable by user or serviceaccount, provider "privileged": Forbidden: not usable by user or serviceaccount]
```
I suspect that the sub chart ran again into that issue we saw a while ago about not being able to set values to null. Looking at the generated stateful set for postgres, it has :

```
securityContext:
        fsGroup: 1001
        seccompProfile:
          type: RuntimeDefault
```

But  for OCP fsGroup should be null: `postgresql.primary.podSecurityContext.fsGroup=null` 

Also, container has:

```
 securityContext:
            allowPrivilegeEscalation: false
            capabilities:
              drop:
              - ALL
            runAsGroup: 0
            runAsNonRoot: true
            runAsUser: 1001
            seccompProfile:
              type: RuntimeDefault
 ```
 
 But    it should have `"postgresql.primary.containerSecurityContext.runAsUser=null"`

pdettori: Only after manually editing the stateful set to delete those two lines (fsGroup: 1001 and runAsUser: 1001), the stateful set started correctly in OCP. So we need to find a solution for nulling values in the sub chart to make this work.

pdettori (changes requested):
Overall this PR would make things cleaner. There are couple comments I made that should be easy to address. But the main concern is that OCP requires some values to be nullified, and that does not seem to work with the sub-chart, likely for a bug in helm (?). We may have to wait for a resolution from community for that unless there is some workaround that can be applied here.

francostellari:
Due to the bug in Helm sub-charts handling, this PR was replaced by #201

PR #185, opened 2024-01-25
✨ Add `--app-version` to the goreleaser packaging action

francostellari:
## Summary

Add `--app-version` to the goreleaser packaging action

## Related issue(s)

Fixes #

### DIFF
--- a/.github/workflows/goreleaser.yml
+++ b/.github/workflows/goreleaser.yml
@@ -8,7 +8,7 @@ on:
 permissions:
   contents: write
   packages: write
- 
+
 env:
   REGISTRY: ghcr.io
   OPERATOR_IMAGE: kubestellar/kubeflex/manager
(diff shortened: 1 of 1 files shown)


.github/workflows/goreleaser.yml
```diff
@@ -61,6 +61,6 @@ jobs:
     - name: Package and push chart
       run: |
         make chart IMG=${{ env.REGISTRY }}/${{ env.OPERATOR_IMAGE }}:${{github.ref_name}}
-        helm package ${{ env.CHART_PATH }} --destination . --version ${{github.ref_name}}
-        helm push ./*.tgz oci://${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}    
+        helm package ${{ env.CHART_PATH }} --destination . --version ${{github.ref_name}} --app-version ${{github.ref_name}} --dependency-update
```
pdettori, line 64: @francostellari do we need the dependency update even if we do not use the sub-chart ?

francostellari:
No but it does not hurt.
I thought to leave it in in case we need it in the future.
If you rather not risk it, I will remove it.

pdettori:
We can leave it, it should not make a difference.

pdettori (approved):
/lgtm

PR #201, opened 2024-02-12
✨ Automatically start postgresql from kubeflex chart

francostellari:
## Summary

- Updated `kubeflex` chart to automatically start `postgresql` dependency using an init container

Additional work needed before merging:
- update Go code for `kflex init`
- update doc

This intends to replace PR #180 which does not work due to Helm bug with sub-charts

## Related issue(s)

Fixes #


pdettori (changes requested):
Thank you for the PR @francostellari. I have a couple of general comments/suggestions:

1. The content in chart/template/operator.yaml is generated from `make chart` which uses Kustomize and the templates under config/ to generate the content for the chart. The current release process does run `make chart` so the content here would be overwritten in the release.

2. Although this works it seems a bit unusual to install the helm chart for the db from an init container. If the pod is restarted (which may happen often as Kube does that) the init container will run every time and will upgrade the release (although the content is the same ) every time.

I think you may use a similar approach with a [helm pre-install hook](https://helm.sh/docs/topics/charts_hooks/). You may have a job that does what this init container does and have the pre-install hook annotation there. This would clearly separate the install of the pre-reqs from the runtime operator.

3. You need also to think about delete - when you delete the kubeflex chart, - with the init container the postgres chart would stay  - nothing removes that. But you might have a post-delete helm hook that runs a job to delete the helm chart.

francostellari:
Can I add other YAML files to the template folder without being overwritten?
`make chart` only overwrites `builtin-hooks.yaml`, `install-hooks.yaml`, `opertor.yaml`.

pdettori:
> Can I add other YAML files to the template folder without being overwritten?

Yes, you can. Only the files you indicated are overwritten.

francostellari:
@pdettori thank you for the comments and suggestions.
I have replaced the initContainer deployment mechanism with Helm hooks to install/uninstall the postgress app.
I have tested it in both Kind and OCP.
No need to install the postgresql chart manually.
This new feature can be desabled by using `--set installPostgreSQL=false`

In kind:
```shell
k create ns kubeflex-system
helm upgrade --install kubeflex-operator chart/ --namespace kubeflex-system --set domain=localtest.me --set externalPort=9443
kflex create cp1
```

in OCP:
```shell
k create ns kubeflex-system
helm upgrade --install kubeflex-operator chart/ --namespace kubeflex-system --set isOpenShift=true
kflex create cp1
```

A pre-delete hook is also provided to delete the posgresql pod automatically when uninstalling kubefelex chart:

```shell
helm uninstall kubeflex-operator --namespace kubeflex-syste
```

pdettori:
@francostellari this looks good. This should be the approach to install KubeFlex with helm, right? For the `kflex init` approach, I would still like to keep the charts install separate and set installPostgreSQL=false - this allows to deploy first postgres and check it is started before starting the kubeflex controller. Bottom line - could you also add to the PR the following:

1. Update the kubeflex user doc for helm, documenting also the use of the flag installPostgreSQL and how to deploy on OpenShift - your update chart would make all these outdated.
https://github.com/kubestellar/kubeflex/blob/main/docs/users.md#installing-kubeflex-with-helm 

2. Update the helm options for deploying the kubeflex helm chart with installPostgreSQL=false in https://github.com/kubestellar/kubeflex/blob/b7372d1c66bf2700bf233c823a59053f0010e350/cmd/kflex/init/init.go#L125-L131

pdettori:
@francostellari it might be better, otherwise we have to ask the user to still run that manual step. Only issue might be about finding an image for the `oc` CLI.

pdettori (approved):
/lgtm

pdettori:
After merging of https://github.com/kubestellar/kubeflex/pull/204 this is now good to merge

PR #207, opened 2024-02-17
🌱 Avoid otel 114

MikeSpreitzer:
## Summary
This PR changes the version of Helm that KubeFlex depends on, from 3.12 to 3.11.2, so that the dependency on Helm does not in turn bring in a dependency on version 1.14.0 of opentelemetry. That dependency is a problem for compiling clients of Kubernetes release 0.28.anything, which depend on opentelemetry 1.10. They made a non-backwards-compatible change from 1.10 to 1.14, removing `WithTLSConfig` from https://github.com/open-telemetry/opentelemetry-go/blob/v1.10.0/exporters/otlp/internal/envconfig/envconfig.go.

## Related issue(s)

Fixes #


MikeSpreitzer:
I can not fix the problem by just go.mod "replace" statements about opentelemetry. I tried that in https://github.com/MikeSpreitzer/kcp-edge-mc/tree/bad-otel-demo, but then something else gets wedged:

```
mspreitz@mjs13 kubestellar % go test ./pkg/binding
# k8s.io/apiserver/pkg/server/routes
../../../../pkg/mod/k8s.io/apiserver@v0.28.2/pkg/server/routes/openapi.go:68:89: cannot use oa.Config (variable of type *common.Config) as *common.OpenAPIV3Config value in argument to builder3.BuildOpenAPISpecFromRoutes
# k8s.io/apiextensions-apiserver/pkg/controller/openapi/builder
../../../../pkg/mod/k8s.io/apiextensions-apiserver@v0.28.2/pkg/controller/openapi/builder/builder.go:204:107: cannot use b.getOpenAPIConfig(false) (value of type *"k8s.io/kube-openapi/pkg/common".Config) as *"k8s.io/kube-openapi/pkg/common".OpenAPIV3Config value in argument to builder3.BuildOpenAPISpecFromRoutes
FAIL	github.com/kubestellar/kubestellar/pkg/binding [build failed]
FAIL
```

MikeSpreitzer:
It was cruft in my indirect requirements.

I do not need this PR; I can get past the opentelemetry botch by adding `replace` statements about opentelemetry in my go.mod.

pdettori:
@MikeSpreitzer ok, so should we close this one for now?

MikeSpreitzer:
No, I think this is still good. It is updating dependencies to remove a problem for clients.

pdettori:
@MikeSpreitzer there is a check that does not pass - see https://github.com/kubestellar/kubeflex/actions/runs/7943445058/job/21687829561?pr=207

can you try to rebase and update the PR?

MikeSpreitzer:
Rebased as suggested and planned.

MikeSpreitzer:
Hmm, now something else is going wrong. And the same thing even strikes commit e8dc40e . Why? Investigating...

MikeSpreitzer:
Release 3.11.2 of Helm uses Kubernetes release 1.26.0. The Kubernetes project is up-front about the fact that they do not follow semver rules; clients of 1.26.0 do not necessarily work with 1.27.0 or higher. We need Kubernetes 1.28.2 currently. So the approach in this PR is doomed.

PR #217, opened 2024-02-22
✨ Switch to `kubestellar` registry

francostellari:
## Summary

Switch to use `quay.io/kubestellar/helm` container image

## Related issue(s)

Fixes #


pdettori (approved):
/lgtm

PR #222, opened 2024-03-01
✨ Update helm chart and docs to allow for the installation of CP, PCH, statusaddon

### LINKED ISSUE #213: feature: Move CP and PCH CRDs out of `operator.yaml` into the special `chart/crds` folder
### Feature Description

To ensure that CRDs are available to be sued by the chart, Helms makes user to install and wait for the avilability of the CRDs that are in the special folder `chart/crds`. This allows other components of the chart to install CRs.

Here is an example of error that may otherwise happen:

```
Error: INSTALLATION FAILED: unable to build kubernetes objects from release manifest: [resource mapping not found for name: "kubestellar" namespace: "" from "": no matches for kind "PostCreateHook" in version "tenancy.kflex.kubestellar.org/v1alpha1"
ensure CRDs are installed first, resource mapping not found for name: "ocm" namespace: "" from "": no matches for kind "PostCreateHook" in version "tenancy.kflex.kubestellar.org/v1alpha1"
ensure CRDs are installed first]
```

### Proposed Solution

Modify the `make chart` process to separate the CP and PCH CRDs into the special folder `chart/crds`

### Want to contribute?

### Additional Context

_No response_

francostellari:
## Summary

This PR intends to update helm chart and docs to allow for the installation of CP, PCH, statusaddon

1. Update the `make chart` to separate the CP and PCH crds from `operator.yaml` to `crds/` folder to guarantee that they are installed before the remaining of the kubeflex chart
2. Update the helm chart to allow for the automatic installation (default off) of `ocm`, `kubestellar`, `statusaddon` Post Create Hooks 
3. Update the helm chart to allow for the automatic installation (default off) of user-defined Control Planes
4. Updated the helm chart `values.yaml` with commented examples of CPs
5. rename `postgresql-hook.yaml` to simply `postgresql.yaml`
7. Update the `user.md` docs to reflect the changes
9. Provide a convenience script to create a suitable kind cluster

All new features are turned off by default.

## Related issue(s)

Fixes #213


scripts/create-kind-cluster.sh
```diff
@@ -0,0 +1,49 @@
...
+    nodeRegistration:
+      kubeletExtraArgs:
+        node-labels: "ingress-ready=true"
+  extraPortMappings:
+  - containerPort: 443
+    hostPort: 9443 # required by kflex
+    protocol: TCP
+EOF
+
+echo Patching nginx ingress with SSL passthrough...
```
MikeSpreitzer, line 36: The command below is a _create_ rather than a _patch_, right?

MikeSpreitzer:
This is creating several references from the kubeflex repo to the kubestellar repo. Yes, some are commented out, but: (a) they are still there and (b) others are not. Dependency cycles between repos are problematic.

Couldn't this work be put in the kubestellar repo instead? I am thinking of leaving a Helm chart in kubeflex that just installs kubeflex, and introducing a Helm chart in kubestellar that (a) has the kubeflex chart as a dependency (listed in `dependencies` in Chart.yaml) and (b) adds all the other specifics. Or, perhaps, extend the Helm chart in kubeflex to add _facilities_ for PCHs that install Helm charts and extend OCM --- without even mentioning kubestellar --- and introduce a Helm chart in kubestellar that depends on the kubeflex Helm chart and uses its new facilities.

ezrasilvera:
Regarding dependency issues KS <-->  KF.  The source of the issue is due to the desire to leave all options open:
1. Space access
    - If KS controllers get the space(s) config directly as parameters -->  KS will not be dependent on KF
    - IF KS controllers retrieve the space config from KF --> KS depends on KF 
2. Deploy controllers into a cluster 
    - If KS controllers are deployed directly into a cluster  -->  KS and KF are not dependent
    - IF KS controllers are deployed through PCH  --> KF depends on KS

francostellari:
@ezrasilvera @MikeSpreitzer can you clarify what cycle dependency do you see?
I'm not sure what do you see in this that is not already done by `kflex`.
I cannot understand https://github.com/kubestellar/kubeflex/pull/222#issuecomment-1986987101
A comment is not a cycle dependency.

scripts/create-kind-cluster.sh
francostellari: easy to fix

scripts/create-kind-cluster.sh
```diff
@@ -0,0 +1,49 @@
...
+
+kubectl create -f https://raw.githubusercontent.com/kubestellar/kubestellar/main/example/kind-nginx-ingress-with-SSL-passthrough.yaml
```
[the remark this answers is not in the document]
francostellari: easy to fix

scripts/create-kind-cluster.sh
francostellari: actually the command is creating something to "patch nginx", so the echo is correct as is

ezrasilvera:
@francostellari I don't understand wat you mean by "A comment is not a cycle dependency."  ..
The dependency today is that the code inside the controllers is using KF in order to bring the space(s) information by fetching the ControlPlane information from KF.  This means that KS depends on KF

ezrasilvera:
@francostellari My last comment on dependency is indeed not related to the Helm dependency . So you can mark it as done. 
Going through the PR I'm not sure you addressed all of @MikeSpreitzer comments.   
@MikeSpreitzer - can you please see @francostellari response at https://github.com/kubestellar/kubeflex/pull/222#issuecomment-1992643816

francostellari:
Team does not think this is a good idea.

MikeSpreitzer:
@francostellari, @pdettori: I think that if KubeFlex is successful then we may want to resurrect this PR because it has a couple of good ideas for general use: making it easy to install CPs and PCHs. My problem with the current draft is that those general facilities are not presented as general facilities but rather things that are here to support ks/ks. If the naming and wording were generalized to be appropriate for any larger system using kubeflex, then I think this would make sense.

PR #231, opened 2024-04-03
✨ Separate CRDs

francostellari:
## Summary

Updated the `Makefile` and kustomizer to separate the CRDs of the helm from the `operator.yaml`

CRDs installed by an helm chart should be included in the special chart/crds/ folder to make sure that Helm installs them first, before acting on the rest of the chart.

## Related issue(s)

Fixes #


ezrasilvera:
@francostellari Can you please add couple of lines to the description about the reason to do this ?

francostellari:
> @francostellari Can you please add couple of lines to the description about the reason to do this ?

@ezrasilvera I have updated the comment as you requested:

CRDs installed by an helm chart should be included in the special chart/crds/ folder to make sure that Helm installs them first, before acting on the rest of the chart.

pdettori (approved):
/lgtm

francostellari:
@pdettori please note that @MikeSpreitzer requested a different approach used for the status-addon: https://github.com/kubestellar/ocm-status-addon/pull/47/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52
Let me know if you want me to match it here

pdettori:
> @pdettori please note that @MikeSpreitzer requested a different approach used for the status-addon: 

@francostellari is the difference about the use of yq ? I am fine both ways, so let's leave this as it is.

francostellari:
@pdettori ok

PR #242, opened 2024-05-02
🐛 Fix postgres install hook role binding

francostellari:
## Summary

When installing the helm chart without specifying the `kubeflex-system` namespace, the `helm-install-rolebinding` is created in the default namespace instead of the `kubeflex-system` namespace. This causes the postgres installation to fail

## Related issue(s)

Fixes #


pdettori (approved):
/lgtm

PR #245, opened 2024-05-03
✨Add initContainer to wait for postgresql Ready status

francostellari:
## Summary

Add initContainer to wait for postgresql Ready status.
This prevent kubeflex from acting on CP CRs before the database is ready
This dramatically speedup the creation of CPs and prevent crashes.

## Related issue(s)

Fixes #

### DIFF
--- a/chart/templates/operator.yaml
+++ b/chart/templates/operator.yaml
@@ -646,6 +646,18 @@ spec:
           capabilities:
             drop:
             - ALL
+      initContainers:
+      - args:
+        - |
+          while [[ $(kubectl get pods postgres-postgresql-0 -n kubeflex-system -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}') != "True" ]]; do
+            echo "waiting for postgresql pod"
+            sleep 5
+          done
+        command:
+        - sh
+        - -c
+        image: quay.io/kubestellar/kubectl:1.29.3
+        name: wait-postgresql
       securityContext:
         runAsNonRoot: true
       serviceAccountName: kubeflex-controller-manager
--- a/config/default/manager_auth_proxy_patch.yaml
+++ b/config/default/manager_auth_proxy_patch.yaml
@@ -24,6 +24,16 @@ spec:
                   operator: In
                   values:
                     - linux
+      initContainers:
+      - name: wait-postgresql
+        image: quay.io/kubestellar/kubectl:1.29.3
+        command: ['sh', '-c']
+        args:
+        - |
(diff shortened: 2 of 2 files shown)


francostellari:
@pdettori (cc @MikeSpreitzer )

config/default/manager_auth_proxy_patch.yaml
```diff
@@ -24,6 +24,16 @@ spec:
...
+      initContainers:
+      - name: wait-postgresql
+        image: quay.io/kubestellar/kubectl:1.29.3
+        command: ['sh', '-c']
+        args:
+        - |
+          while [[ $(kubectl get pods postgres-postgresql-0 -n kubeflex-system -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}') != "True" ]]; do
+            echo "waiting for postgresql pod"
+            sleep 5
+          done
```
MikeSpreitzer, line 36: Why does this need to be here?

chart/templates/operator.yaml
```diff
@@ -646,6 +646,18 @@ spec:
           capabilities:
             drop:
             - ALL
+      initContainers:
+      - args:
+        - |
```
MikeSpreitzer, line 651: Why not include this arg along with the other one in `command`?

chart/templates/operator.yaml
```diff
@@ -646,6 +646,18 @@ spec:
...
+          while [[ $(kubectl get pods postgres-postgresql-0 -n kubeflex-system -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}') != "True" ]]; do
```
MikeSpreitzer, line 652: What happens if the nested command `kubectl get pods postgres-postgresql-0 -n kubeflex-system -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}'` fails, in the sense of exiting with a non-zero status? Will that make the `[[ ... ]]` exit with a non-zero status?

I think that the following is more clearly safe.

```
while ! [[ $(kubectl get pods postgres-postgresql-0 -n kubeflex-system -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}') == "True" ]]; do
```

config/default/manager_auth_proxy_patch.yaml
francostellari: @MikeSpreitzer because `operator.yaml' is generated from this file by `make chart`.
Changes made in 'operator.yaml` would be overridden by kustomizer.

chart/templates/operator.yaml
francostellari: I guess it is a preference, I like to write the script code on multiple lines, easier to copy&past for testing

pdettori (approved):
/lgtm

PR #249, opened 2024-05-15
✨Enable verbosity control from chart values

francostellari:
## Summary

Enable verbosity control from chart values

## Related issue(s)

Fixes #


pdettori (approved):
/lgtm

Issue #195, opened 2024-02-08
bug:  Command `kflex create wds1 -p kubestellar` fails because a security issue in Microshift 4.8


francostellari:
### Describe the bug

Command `kflex create wds1 -p kubestellar` fails in Microshift 4.8

```
wds1-system                     kube-apiserver-7f9478566d-sn2lg                                   0/2     CreateContainerConfigError   0          107s
```

```
Events:
  Type     Reason     Age                   From               Message
  ----     ------     ----                  ----               -------
  Normal   Scheduled  3m9s                  default-scheduler  Successfully assigned wds1-system/kube-apiserver-7f9478566d-sn2lg to pica-spi.watson.ibm.com
  Normal   Pulling    3m8s                  kubelet            Pulling image "rancher/kine:v0.11.4"
  Normal   Pulled     3m                    kubelet            Successfully pulled image "rancher/kine:v0.11.4" in 7.639057739s
  Normal   Pulling    3m                    kubelet            Pulling image "registry.k8s.io/kube-apiserver:v1.27.1"
  Normal   Pulled     2m41s                 kubelet            Successfully pulled image "registry.k8s.io/kube-apiserver:v1.27.1" in 18.349763951s
  Warning  Failed     119s (x5 over 2m41s)  kubelet            Error: container has runAsNonRoot and image will run as root (pod: "kube-apiserver-7f9478566d-sn2lg_wds1-system(8977c5a5-4d40-48ed-8b1b-4123bbaa4c93)", container: kube-apiserver)
  Warning  Failed     106s (x6 over 3m)     kubelet            Error: container has runAsNonRoot and image has non-numeric user (nobody), cannot verify user is non-root (pod: "kube-apiserver-7f9478566d-sn2lg_wds1-system(8977c5a5-4d40-48ed-8b1b-4123bbaa4c93)", container: kine)
  Normal   Pulled     106s (x5 over 2m41s)  kubelet            Container image "rancher/kine:v0.11.4" already present on machine
  Normal   Pulled     106s (x5 over 2m41s)  kubelet            Container image "registry.k8s.io/kube-apiserver:v1.27.1" already present on machine
```

### Steps To Reproduce

Run `kflex create wds1 -p kubestellar` in Microshift 4.8 and observe that it fails because of a security issue

### Expected Behavior

The pod should start correctly

### Additional Context

The issue seems to be fixed by manually editing the deployment yaml and adding `runAsUser: 10001` to both containers securityContext

[...]

PR #259, opened 2024-05-23
🐛 Enable the use of KubeFlex in MicroShift cluster

francostellari:
## Summary

Add a new `isMicroShift` chart value to let the controller know that when creating a `k8s` control plane in a MicroShift cluster a `runAsUser` value is needed.

## Related issue(s)

Fixes #
#195

### DIFF
--- a/pkg/reconcilers/k8s/deployment.go
+++ b/pkg/reconcilers/k8s/deployment.go
@@ -36,7 +36,7 @@ import (
 	"github.com/kubestellar/kubeflex/pkg/util"
 )
 
-func (r *K8sReconciler) ReconcileAPIServerDeployment(ctx context.Context, hcp *tenancyv1alpha1.ControlPlane, isOCP bool) error {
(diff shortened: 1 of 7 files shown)


pkg/reconcilers/k8s/deployment.go
```diff
@@ -266,7 +266,16 @@ func (r *K8sReconciler) generateAPIServerDeployment(namespace, dbName string, is
 			},
 		},
 	}
-	if isOCP {
+	if isMicroShift {
+		deployment.Spec.Template.Spec.SecurityContext = &v1.PodSecurityContext{
+			RunAsNonRoot: pointer.Bool(true),
+			RunAsUser:    pointer.Int64(10001),
```
MikeSpreitzer, line 272: I thought we used 65534 in other fixes. 65534 is pretty standard, usually having username "nobody". AFAIK, 10001 is not so standardized.

francostellari:
ok

PR #263, opened 2024-05-24
🐛 fix setting missing context for cp of type host

### LINKED ISSUE #260: bug: Unable to `kflex ctx` to control plane of type "host"
### Describe the bug

After using the core helm chart to deploy a kubestellar core control plane with its1 configured as type=host, there is no way to access an its1 context or create an its1 context with kflex. 

### Steps To Reproduce

1. on a VM or bare metal machine, install k3s
```
sudo curl -sfL https://get.k3s.io | INSTALL_K3S_EXEC="--disable traefik,servicelb"  sh -
export KUBECONFIG=~/.kube/config
sudo kubectl --kubeconfig=/etc/rancher/k3s/k3s.yaml config view --raw > "$KUBECONFIG"
```
2. install ingress nginx
```
helm upgrade --install ingress-nginx ingress-nginx --repo https://kubernetes.github.io/ingress-nginx --namespace ingress-nginx --create-namespace \
       --set controller.updateStrategy.type=RollingUpdate --set controller.updateStrategy.rollingUpdate.maxUnavailable=1 --set controller.service.type=NodePort \
       --set controller.publishService.enabled=false --set controller.hostPort.enabled=true --set controller.hostPort.ports.https=9443 \
       --set controller.extraArgs.enable-ssl-passthrough=true
```
3. Install core chart with wds1 and its1 of type=host
`helm upgrade --install ks-hub oci://ghcr.io/francostellari/kubestellar/hub-chart --version 0.1.2  --set-json='ITSes=[{"name":"its1","type":"host"}]'  --set-json='WDSes=[{"name":"wds1"}]'`

4. Try to create contexts for wds1 and its1
`for cpname in `kubectl get controlplane -o name`; do cpname=${cpname##*/};   echo "Getting the kubeconfig of Control Planes \"$cpname\"...";   kflex ctx $cpname; done`

### Expected Behavior

I would expect that given a valid control plane ($cpname) the command `kflex ctx $cpname` would create the kubectl context if necessary.

### Additional Context

_No response_

pdettori:
## Summary

fix setting missing context for cp of type host

## Related issue(s)

Fixes #260

### DIFF
--- a/cmd/kflex/ctx/ctx.go
+++ b/cmd/kflex/ctx/ctx.go
@@ -92,8 +92,8 @@ func (c *CPCtx) Context(chattyStatus, failIfNone bool) {
 					os.Exit(1)
 				}
 			} else {
-				fmt.Fprintf(os.Stderr, "control plane %s is of type 'host', cannot switch context\n", c.Name)
-				os.Exit(1)
+				fmt.Fprintf(os.Stderr, "control plane %s is of type 'host', using hosting cluster context\n", c.Name)
+				os.Exit(0)
 			}
 		}
 		done <- true
--- a/cmd/kflex/ctx/ctx.go
+++ b/cmd/kflex/ctx/ctx.go
@@ -36,6 +36,7 @@ import (
 
 type CPCtx struct {
 	common.CP
+	Type tenancyv1alpha1.ControlPlaneType
 }
 
 // Context switch context in Kubeconfig
@@ -84,8 +85,14 @@ func (c *CPCtx) Context(chattyStatus, failIfNone bool) {
 				fmt.Fprintf(os.Stderr, "Error loading kubeconfig context from server: %s\n", err)
 				os.Exit(1)
 			}
-			if err = kubeconfig.SwitchContext(kconf, c.Name); err != nil {
-				fmt.Fprintf(os.Stderr, "Error switching kubeconfig context after loading from server: %s\n", err)
+			// context exists only for CPs that are not of type host
+			if c.Type != tenancyv1alpha1.ControlPlaneTypeHost {
+				if err = kubeconfig.SwitchContext(kconf, c.Name); err != nil {
+					fmt.Fprintf(os.Stderr, "Error switching kubeconfig context after loading from server: %s\n", err)
+					os.Exit(1)
+				}
+			} else {
+				fmt.Fprintf(os.Stderr, "control plane %s is of type 'host', cannot switch context\n", c.Name)
 				os.Exit(1)
 			}
 		}
@@ -115,15 +122,11 @@ func (c *CPCtx) loadAndMergeFromServer(kconfig *api.Config) error {
 	if err := kfcClient.Get(context.TODO(), client.ObjectKeyFromObject(cp), cp, &client.GetOptions{}); err != nil {
 		return fmt.Errorf("control plane not found on server: %s", err)
 	}
+	c.Type = cp.Spec.Type
 
-	// for control plane of type host just create context alias and return
+	// for control plane of type host just switch to initial context
 	if cp.Spec.Type == tenancyv1alpha1.ControlPlaneTypeHost {
-		err = kubeconfig.CopyHostContextAndSetItToDefault(kconfig, c.Name)
-		if err != nil {
-			fmt.Fprintf(os.Stderr, "error copying hosting cluster context and setting it as default: %s\n", err)
-			os.Exit(1)
-		}
-		return nil
+		return kubeconfig.SwitchToInitialContext(kconfig, false)
 	}
 
 	// for all other control planes need to get secret with off-cluster kubeconfig
--- a/cmd/kflex/ctx/ctx.go
+++ b/cmd/kflex/ctx/ctx.go
@@ -76,7 +76,7 @@ func (c *CPCtx) Context(chattyStatus, failIfNone bool) {
 		ctxName := certs.GenerateContextName(c.Name)
 		util.PrintStatus(fmt.Sprintf("Switching to context %s...", ctxName), done, &wg, chattyStatus)
 		if err = kubeconfig.SwitchContext(kconf, c.Name); err != nil {
-			fmt.Fprintf(os.Stderr, "kubeconfig context %s not found, trying to load from server...\n", err)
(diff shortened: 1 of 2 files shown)


cmd/kflex/ctx/ctx.go
```diff
@@ -106,6 +106,17 @@ func (c *CPCtx) loadAndMergeFromServer(kconfig *api.Config) error {
 		return fmt.Errorf("control plane not found on server: %s", err)
 	}
 
+	// for control plane of type host just create context alias and return
+	if cp.Spec.Type == tenancyv1alpha1.ControlPlaneTypeHost {
+		err = kubeconfig.CopyHostContextAndSetItToDefault(kconfig, c.Name)
```
MikeSpreitzer, line 111: Why make a copy? Why not use the original?

pdettori: The copy of the context or "alias" context is part of the current UX, and it is done mainly for a consistent experience - the idea is that there is a context for each control plane with the same name of the control plane, independently of the type of control plane.

pdettori: @MikeSpreitzer, let me see if I understand the proposal - if the control plane is of type host the idea not to create an alias context (as a copy of the hosting cluster context) with the name of the control plane, but just use the hosting cluster context. That is, `kflex ctx <cp-of-type-host>` is equivalent to `kflex ctx`.  That may make possibly simpler understanding what is going on when using a cp of type host. Is that correct ?

MikeSpreitzer: yes.
Along with making `kflex ctx` set the hosting context name extension IF it is not already set and the current context IS the hosting context.

pdettori:
@MikeSpreitzer removed use of alias for host context as suggested

cmd/kflex/ctx/ctx.go
```diff
@@ -79,8 +80,14 @@ func (c *CPCtx) Context(chattyStatus, failIfNone bool) {
...
-			if err = kubeconfig.SwitchContext(kconf, c.Name); err != nil {
-				fmt.Fprintf(os.Stderr, "Error switching kubeconfig context after loading from server: %s\n", err)
+			// context exists only for CPs that are not of type host
+			if c.Type != tenancyv1alpha1.ControlPlaneTypeHost {
+				if err = kubeconfig.SwitchContext(kconf, c.Name); err != nil {
+					fmt.Fprintf(os.Stderr, "Error switching kubeconfig context after loading from server: %s\n", err)
+					os.Exit(1)
+				}
+			} else {
+				fmt.Fprintf(os.Stderr, "control plane %s is of type 'host', cannot switch context\n", c.Name)
```
MikeSpreitzer, line 90: I would not characterize this as a failure; the context has been successfully switched to the context of the requested control plane.

cmd/kflex/ctx/ctx.go
```diff
@@ -79,8 +80,14 @@ func (c *CPCtx) Context(chattyStatus, failIfNone bool) {
...
 				os.Exit(1)
```
MikeSpreitzer, line 91: I would not characterize this as a failure; the context has been successfully switched to the context of the given control plane.

cmd/kflex/ctx/ctx.go
```diff
@@ -84,9 +85,15 @@ func (c *CPCtx) Context(chattyStatus, failIfNone bool) {
...
-				fmt.Fprintf(os.Stderr, "Error switching kubeconfig context after loading from server: %s\n", err)
-				os.Exit(1)
+			// context exists only for CPs that are not of type host
+			if c.Type != tenancyv1alpha1.ControlPlaneTypeHost {
+				if err = kubeconfig.SwitchContext(kconf, c.Name); err != nil {
+					fmt.Fprintf(os.Stderr, "Error switching kubeconfig context after loading from server: %s\n", err)
+					os.Exit(1)
+				}
+			} else {
+				fmt.Fprintf(os.Stderr, "control plane %s is of type 'host', using hosting cluster context\n", c.Name)
```
MikeSpreitzer, line 95: This is good. Outputting the name of the hosting cluster context would be better. Does not have to be changed in this PR.

cmd/kflex/ctx/ctx.go
```diff
@@ -84,9 +85,15 @@ func (c *CPCtx) Context(chattyStatus, failIfNone bool) {
...
+				os.Exit(0)
```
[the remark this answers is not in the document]
pdettori: This is what happens if I write to the done channel when switching context for cp of type host (cp2)

```
3.10.0) (⎈|cp1:default)$ kflex ctx cp2
kubeconfig context context cp2 not found not found, trying to load from server...
◐ Switching to context cp2...control plane cp2 is of type 'host', using hosting cluster context
✔ Switching to context cp2...
(3.10.0) (⎈|kind-kubeflex:N/A)$ 
```

So it seems like switching context to cp2 was successful, but context did not switch to cp2 - the "switching to context" message was printed at the beginning, before the type of CP was determined.

This is the behavior instead when I have `os.Exit(0)`

```
(3.10.0) (⎈|cp1:default)$ kflex ctx cp2
kubeconfig context context cp2 not found not found, trying to load from server...
◐ Switching to context cp2...control plane cp2 is of type 'host', using hosting cluster context
```

cmd/kflex/ctx/ctx.go
```diff
@@ -75,7 +76,7 @@ func (c *CPCtx) Context(chattyStatus, failIfNone bool) {
 		ctxName := certs.GenerateContextName(c.Name)
 		util.PrintStatus(fmt.Sprintf("Switching to context %s...", ctxName), done, &wg, chattyStatus)
 		if err = kubeconfig.SwitchContext(kconf, c.Name); err != nil {
-			fmt.Fprintf(os.Stderr, "kubeconfig context %s not found, trying to load from server...\n", err)
+			fmt.Fprintf(os.Stderr, "kubeconfig context %s not found, trying to load from server...\n", c.Name)
```
MikeSpreitzer, line 79: It wouldn't hurt to print the error, as a parenthetic addition.

MikeSpreitzer (commented):
/lgtm
Thanks!

MikeSpreitzer (approved):
Looks good to me.

PR #265, opened 2024-05-24
🐛 wait for control plane kubeconfig secret to be created

### LINKED ISSUE #258: `kflex ctx $cpname` is unnecessarily impatient
If a ControlPlane object named `$foo` has been created but the corresponding Secret has not yet been created, `kflex ctx $foo` will fail. It would be better to wait instead.

pdettori:
## Summary

 wait for control plane kubeconfig secret to be created

## Related issue(s)

Fixes #258


MikeSpreitzer (approved):
Looks good to me.

PR #266, opened 2024-05-24
🐛 set hosting cluster context preference in kflex init

### LINKED ISSUE #255: kflex storage of hosting cluster context name is unnecessarily fragile
1. `kflex init` does NOT store the name of the hosting context. That is done later when `kflex create` or `kflex ctx $cpname` is used to switch away. This opens an unnecessary window for problems: the user might (unknowingly or inadvertently) use `kubectl config use-context` for some reason before doing `kflex create`.
2. If the name of the hosting context is not currently stored in the user's kubeconfig file, and the current context is the hosting cluster context, `kflex ctx` does not store the context name. This is unnecessarily missing an opportunity to recover needed information.

pdettori:
## Summary

- set hosting cluster context preference in kflex init
- use "hosting cluster context" instead of "initial context" to make more clear the role of such context

## Related issue(s)

Fixes #255


pdettori:
Need to merge #265 first

MikeSpreitzer:
@pdettori:

> and the current context is the hosting cluster context

which can be tested by looking to see if the controlplanes resource exists.

pdettori:
@MikeSpreitzer let me see if I understand your proposal - the idea is not to add a new command to set the hosting cluster context preference (in case is not set or wrong) but rather use `kflex ctx` to do that, by testing if indeed the current context has the `controlplanes` API resource and if yes, set the hosting cluster context preference to the current context, if not, the error message should guide the user to use `kubectl config use-context` to first set the context to the hosting cluster context and then run `kflex ctx` again.

MikeSpreitzer:
@pdettori: yes

MikeSpreitzer:
I think what I am saying is the natural confluence of two things that we already are telling users to expect: (A) `kflex ctx` sets the kubeconfig current context to the hosting cluster context and (B) if the kubeconfig extension is not present but the current context IS the hosting context then the kubeconfig extension will be written (this was already the case for `kflex ctx $my_non_host_cp_name`).

pdettori:
@MikeSpreitzer point 2 in #255 has been addressed

cmd/kflex/init/init.go
```diff
@@ -47,6 +48,14 @@ func Init(ctx context.Context, kubeconfig, version, buildDate string, domain, ex
 	util.PrintStatus(fmt.Sprintf("Kubeflex %s %s", version, buildDate), done, &wg, chattyStatus)
 	done <- true
 
+	util.PrintStatus("Setting hosting cluster preferences in kubeconfig", done, &wg, chattyStatus)
```
MikeSpreitzer, line 51: It is a singular preference, right?

MikeSpreitzer (commented):
Just one typo, that could be fixed in this PR or later.
/lgtm
Thanks!

pdettori:
Thank you, fixed the type, will go ahead and merge now.

PR #289, opened 2024-10-21
✨ more forceful kflex ctx commands

### LINKED ISSUE #288: feature: Need more forceful kflex commands
### Feature Description

(A) `kflex ctx $something` will do nothing if there is already a kubeconfig context named `$something` --- even if that context is a broken old thing rather than what this command would create if that context did not exist. There should be something like `fklex ctx --force $something` that will replace the context even if it already exists.

(B) `kflex ctx` will not replace it if the kubeconfig extension naming the hosting cluster context exists but holds a bad value (the name of a context that exists but is no longer the right context). There should be something like `kflex ctx use-current-for-hosting` that sets the extension contents unconditionally.

### Proposed Solution

See above.

### Want to contribute?

### Additional Context

_No response_

pdettori:
## Summary

Provides new flags for  the`kflex ctx` command:
- `--overwrite-existing-context`
- `--set-current-for-hosting`
to help with those cases where hosting cluster context stored in kubeconfig is invalid.

Fixes #288

### DIFF
--- a/cmd/kflex/ctx/ctx.go
+++ b/cmd/kflex/ctx/ctx.go
@@ -40,7 +40,7 @@ type CPCtx struct {
 }
 
 // Context switch context in Kubeconfig
-func (c *CPCtx) Context(chattyStatus, failIfNone, forceCtxAsHosting, setCurrentCtxAsHosting bool) {
+func (c *CPCtx) Context(chattyStatus, failIfNone, overwriteExistingCtx, setCurrentCtxAsHosting bool) {
 	done := make(chan bool)
 	var wg sync.WaitGroup
 	kconf, err := kubeconfig.LoadKubeconfig(c.Ctx)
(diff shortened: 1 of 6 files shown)


cmd/kflex/ctx/ctx.go
```diff
@@ -64,39 +67,46 @@ func (c *CPCtx) Context(chattyStatus, failIfNone bool) {
...
-			}
-			if err = c.loadAndMergeFromServer(kconf); err != nil {
-				fmt.Fprintf(os.Stderr, "Error loading kubeconfig context from server: %s\n", err)
-				os.Exit(1)
-			}
-			// context exists only for CPs that are not of type host
-			if c.Type != tenancyv1alpha1.ControlPlaneTypeHost {
-				if err = kubeconfig.SwitchContext(kconf, c.Name); err != nil {
-					fmt.Fprintf(os.Stderr, "Error switching kubeconfig context after loading from server: %s\n", err)
+		// handle the force hostig cluster flag first
```
MikeSpreitzer, line 79: s/hostig/hosting/

cmd/kflex/ctx/ctx.go
```diff
@@ -64,39 +67,46 @@ func (c *CPCtx) Context(chattyStatus, failIfNone bool) {
 		} else if failIfNone {
 			if !c.isCurrentContextHostingClusterContext() {
 				fmt.Fprintln(os.Stderr, "The hosting cluster context is not known!\n"+
-					"You can make it known to kflex by doing `kubectl config use-context $name_of_hosting_context` "+
-					"and then either `kflex ctx` or `kflex create ...`")
+					"You can make it known to kflex by doing `kubectl ctx --force-replace-for-hosting $name_of_hosting_context` "+
```
MikeSpreitzer, line 70: (0) This PR should include a doc update.

(1) Note that Issue #288 calls out two _different_ issues. This PR addresses the one labeled (B), in two ways. I think that the more succinct one would be sufficient. The wordier one, `kflex ctx --force-replace-for-hosting $name_of_hosting_context`, is inconsistent with the normal syntax of `kflex ctx $control_plane_name`.

pdettori: @MikeSpreitzer thanks, so is the suggestion then that only `--set-current-for-hosting` is sufficient?  I am not sure I understood then the original suggestion in #288 (A)  then "There should be something like kflex ctx --force $something that will replace the context even if it already exists." - could you pls. elaborate more on what your suggestion for (A) is?

MikeSpreitzer: Item (B) in Issue 288 concerns what happens when my kubeconfig file contains a context named "foo" but that is outdated or otherwise incorrect. I want to be able to make `kflex ctx foo` _set_ context "foo" to be the right stuff for the control plane "foo" regardless of whether I already had a context named "foo".

MikeSpreitzer: This is so that I can simplify our instructions by replacing

```shell
kubectl config delete-context wds1 || true
kflex ctx wds1
```

with

```shell
kflex ctx --ignore-existing-context wds1
```
(or whatever is a good name for the flag)

pdettori: I see, thanks, so is this about replacing the all the elements associated with a context (context name, clusters, auth) ?

MikeSpreitzer: Right. The idea is to not assume that just because there is a context with the right name that means that it has the right contents of any form. It might be quite old and outdated.

MikeSpreitzer (approved):
Tested successfully.
Looks good to me.

PR #302, opened 2024-11-25
✨  Kube1.29

MikeSpreitzer:
## Summary
This PR raises the minimum Kubernetes release to 1.29, and also raises other minimums.

This PR is built on #301 and I will rebase this one if that one merges first.

## Related issue(s)

Fixes #


go.mod
```diff
@@ -1,79 +1,82 @@
 module github.com/kubestellar/kubeflex
 
-go 1.19
+go 1.22.7
```
MikeSpreitzer, line 3: I did not want to add a patch level to this version string, but `go vet ./...` complains that `go mod tidy` is needed if there is no patch level. I picked 7 because that is the last one for which the release notes say there is a security fix.

pdettori: I see - works for me.

PR #309, opened 2025-01-20
🐛 enqueue secrets created by vcluster in addition to owned secrets

pdettori:
## Summary

enqueue secrets created by vcluster in addition to owned secrets 

Fixes #2717

### DIFF
--- a/internal/controller/controlplane_controller.go
+++ b/internal/controller/controlplane_controller.go
@@ -25,13 +25,15 @@ import (
 	networkingv1 "k8s.io/api/networking/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	ctrl "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	clog "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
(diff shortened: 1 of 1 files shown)


MikeSpreitzer:
This seems a bit broader than necessary. Can there be filtering that narrows down the set of Secrets that the KubeFlex controller will react to?

internal/controller/controlplane_controller.go
```diff
@@ -163,16 +166,38 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
 func (r *ControlPlaneReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&tenancyv1alpha1.ControlPlane{}).
-		Owns(&corev1.Service{}).
 		Owns(&networkingv1.Ingress{}).
 		Owns(&appsv1.Deployment{}).
 		Owns(&appsv1.StatefulSet{}).
```
MikeSpreitzer, line 171: Same problem here too, if I understand https://github.com/kubestellar/kubeflex/blob/v0.7.2/pkg/reconcilers/vcluster/reconciler.go#L100 correctly. If the StatefulSet is missing, this controller will note that in the ControlPlane's status and happily be done, without getting notified when the StatefulSet is created later.

internal/controller/controlplane_controller.go
```diff
@@ -163,16 +166,38 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
-		Owns(&corev1.Secret{}).
 		Owns(&corev1.ConfigMap{}).
 		Owns(&corev1.ServiceAccount{}).
+		Owns(&corev1.Secret{}).
+		Watches(&corev1.Secret{}, enqueueSecretsOfInterest()).
 		Complete(r)
```
MikeSpreitzer, line 176: Also, I see that https://github.com/kubestellar/kubeflex/blob/v0.7.2/pkg/reconcilers/vcluster/reconciler.go#L121 is sensitive to the existence of a PostCreateHook. The usual Kubernetes style is to wait for a referenced object to exist when told to use it.

MikeSpreitzer (approved):
LGTM

This is progress. Further issues noted in independent comments here.

PR #312, opened 2025-02-26
✨ Use postgresql chart and image from quay.io

francostellari:
## Summary

KubeFlex chart was changed to use PostgreSQL chart stored in Quay.io: `oci://quay.io/kubestellar/charts/postgresq:13.1.5`

The chart is a copy of the original `oci://registry-1.docker.io/bitnamicharts/postgresql:13.1.5` but pointing to the container image in quay at `quay.io/kubestellar/postgresql/16.0.0-debian-11-r13` which is a multi-arch (amd64/arm64) copy of the original image at `docker.io/bitnami/postgresql:16.0.0-debian-11-r13`.

The main objective of this PR is to avoid hitting Docker pull request limit, especially during testing.

## Related issue(s)

Fixes #

### DIFF
--- a/chart/templates/postgresql.yaml
+++ b/chart/templates/postgresql.yaml
@@ -94,7 +94,7 @@ spec:
           - upgrade
           - --install
           - postgres
-          - oci://registry-1.docker.io/bitnamicharts/postgresql
+          - oci://quay.io/kubestellar/charts/postgresql
           - --version
           - 13.1.5
           - --namespace


francostellari:
Testing the local chart after `make chart'

```shell
$ k create ns kubeflex-system
$ helm install kf chart/
$ k describe pod postgres-postgresql-0 -n kubeflex-system
....
  Normal  Pulling    31s   kubelet            Pulling image "quay.io/kubestellar/postgresql:16.0.0-debian-11-r13"
  Normal  Pulled     17s   kubelet            Successfully pulled image "quay.io/kubestellar/postgresql:16.0.0-debian-11-r13" in 14.026s (14.026s including waiting). Image size: 91853132 bytes.
...
```

francostellari:
If using the quay image/chart for all users permanently is not acceptable, then I propose to bring out to the `values.yaml` a value for the postgresql chart registry:

```yaml
postgresql:
  image: oci://registry-1.docker.io/bitnamicharts/postgresql
  tag: 13.1.5
```

so that it can be overridden by a user during testing

pdettori (approved):
/lgtm

PR #315, opened 2025-03-03
🐛 remove remaining dockerhub dependencies

pdettori:
## Summary

Removes remaining docker hub deps


MikeSpreitzer (commented):
I have not tested it. I assume that the author has.
LGTM

pdettori:
> I have not tested it. I assume that the author has.

Yes, I have tested it. The e2e tests also exercise the change.

PR #323, opened 2025-03-10
🐛  Fix sharing, enqueuing, and logging bugs

### LINKED ISSUE #318: bug: Wrong ControlPlane reference derived from Secret
### Describe the bug

PR #309 is buggy, it computes the wrong reference to a ControlPlane. It takes the Secret's name and namespace to be that of the ControlPlane, which is simply not true. This causes, for example, log lines like the following.

```
2025-03-07T10:15:10Z	INFO	Got ControlPlane event!	{"controller": "controlplane", "controllerGroup": "tenancy.kflex.kubestellar.org", "controllerKind": "ControlPlane", "ControlPlane": {"name":"vc-vcluster","namespace":"its1-system"}, "namespace": "its1-system", "name": "vc-vcluster", "reconcileID": "3f27bda4-c609-4e01-9e35-bfea523ea68f"}
```

This is the current cause of https://github.com/kubestellar/kubestellar/issues/2717

### Steps To Reproduce

Create a ControlPlane of type "vcluster" and look at the log from the controller.

### Expected Behavior

Correct derivation of ControlPlane reference from Secret.
The vc-vcluster Secret gets its config-incluster data member promptly.

### Additional Context

_No response_

### LINKED ISSUE #321: bug: logging gaps in the controller
### Describe the bug

The following lines silently ignore an error that should not happen. Since this condition stops normal processing, it should be logged in the ControlPlane's status like any other thing that stopped normal processing.

https://github.com/kubestellar/kubeflex/blob/v0.8.1/pkg/reconcilers/shared/ingress.go#L63
https://github.com/kubestellar/kubeflex/blob/v0.8.1/pkg/reconcilers/shared/namespace.go#L48
https://github.com/kubestellar/kubeflex/blob/v0.8.1/pkg/reconcilers/shared/route.go#L56
https://github.com/kubestellar/kubeflex/blob/v0.8.1/pkg/reconcilers/vcluster/service.go#L55
https://github.com/kubestellar/kubeflex/blob/v0.8.1/pkg/reconcilers/shared/rbac.go#L54
https://github.com/kubestellar/kubeflex/blob/v0.8.1/pkg/reconcilers/shared/rbac.go#L109
https://github.com/kubestellar/kubeflex/blob/v0.8.1/pkg/reconcilers/shared/job.go#L61

The following line does not log the namespace that the object will be created in (the object has not had the namespace set).

https://github.com/kubestellar/kubeflex/blob/v0.8.1/pkg/reconcilers/shared/postcreate_hook.go#L136

### Steps To Reproduce

These are from code inspection. The last one is confirmed by log inspection. For example:

```
2025-03-07T10:14:42Z	INFO	Applying	{"controller": "controlplane", "controllerGroup": "tenancy.kflex.kubestellar.org", "controllerKind": "ControlPlane", "ControlPlane": {"name":"wds1"}, "namespace": "", "name": "wds1", "reconcileID": "79848534-5a0d-429e-9741-492aa451210c", "object": "[] deployment.apps/kubestellar-controller-manager"}
```

### Expected Behavior

Errors should not be silently discarded.
The created object namespace should be logged.

### Additional Context

_No response_

### LINKED ISSUE #322: bug: Improper use of a global variable in controller
### Describe the bug

https://github.com/kubestellar/kubeflex/blob/v0.8.1/pkg/reconcilers/vcluster/chart.go re-uses a global variable named `configs` for each of the vcluster ControlPlanes. This is a functional problem and a concurrency bug.

### Steps To Reproduce

This is from code inspection.

### Expected Behavior

Each vcluster's chart is independent. No concurrency bugs.

### Additional Context

_No response_

MikeSpreitzer:
## Summary
This PR fixes three issues.

1. Wrong reference to ControlPlane derived from vc-vcluster Secret.
2. Misuse of a global variable.
3. Logging gaps/bugs.

This PR also adds Event logging for ControlPlane sync results, which should be helpful in the case of transient syncing problems. Issue #318 has a transient sync failure that lasts for a highly variable amount of time. In particular, sometimes it is long enough to cause https://github.com/kubestellar/kubestellar/issues/2717 --- and yet be gone by the time someone investigates.

## Related issue(s)

Fixes #318 
Fixes #322 
Fixes #321

### DIFF
--- a/internal/controller/controlplane_controller.go
+++ b/internal/controller/controlplane_controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/go-logr/logr"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	networkingv1 "k8s.io/api/networking/v1"
@@ -176,11 +177,11 @@ func (r *ControlPlaneReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&corev1.ConfigMap{}).
 		Owns(&corev1.ServiceAccount{}).
 		Owns(&corev1.Secret{}).
-		Watches(&corev1.Secret{}, enqueueSecretsOfInterest()).
+		Watches(&corev1.Secret{}, enqueueSecretsOfInterest(mgr.GetLogger())).
 		Complete(r)
 }
 
-func enqueueSecretsOfInterest() handler.EventHandler {
+func enqueueSecretsOfInterest(logger logr.Logger) handler.EventHandler {
 	return handler.EnqueueRequestsFromMapFunc(func(ctx context.Context, o client.Object) []reconcile.Request {
 		secret, ok := o.(*corev1.Secret)
 		if !ok {
@@ -188,17 +189,15 @@ func enqueueSecretsOfInterest() handler.EventHandler {
 		}
 
 		// Check if the secret has the specified name
-		var matches []reconcile.Request
-		if secret.Name == util.VClusterKubeConfigSecret {
-			matches = append(matches, reconcile.Request{
-				NamespacedName: types.NamespacedName{
-					Name:      secret.GetName(),
-					Namespace: secret.GetNamespace(),
-				},
-			})
+		if secret.Name != util.VClusterKubeConfigSecret {
+			cpName, err := util.ControlPlaneNameFromNamespace(secret.Namespace)
+			if err == nil {
+				return []reconcile.Request{{NamespacedName: types.NamespacedName{Name: cpName}}}
+			}
+			logger.Info("Ignoring non-ControlPlane Secret named "+util.VClusterKubeConfigSecret, "namespace", secret.Namespace)
 		}
 
-		return matches
+		return []reconcile.Request{}
 	})
 }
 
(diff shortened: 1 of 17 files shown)


MikeSpreitzer:
@francostellari: Perhaps the snapshot script should capture Events too.

internal/controller/controlplane_controller.go
```diff
@@ -176,29 +180,27 @@ func (r *ControlPlaneReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-					Name:      secret.GetName(),
-					Namespace: secret.GetNamespace(),
-				},
-			})
+		if secret.Name != util.VClusterKubeConfigSecret {
+			cpName, err := util.ControlPlaneNameFromNamespace(secret.Namespace)
+			if err == nil {
+				return []reconcile.Request{{NamespacedName: types.NamespacedName{Name: cpName}}}
+			}
+			logger.Info("Ignoring non-ControlPlane Secret named "+util.VClusterKubeConfigSecret, "namespace", secret.Namespace)
```
pdettori, line 200: I may be missing something here, but it looks like there is no reconciliation if secret.Name == util.VClusterKubeConfigSecret - I thought that we wanted the controller to react when that secret was created, no ?

MikeSpreitzer: Fixed.

Issue #300, opened 2024-11-01
bug: wait for condition not working as expected with control plane conditions


pdettori:
### Describe the bug

(from @MikeSpreitzer )

These two are supposed to do the same thing. How do they not?
```
mspreitz@mjs13 kubestellar % kubectl --context kind-kubeflex wait controlplane/its1 --for 'jsonpath={.status.conditions[?(@.type=="Synced")].status}=True'
controlplane.tenancy.kflex.kubestellar.org/its1 condition met

mspreitz@mjs13 kubestellar % kubectl --context kind-kubeflex wait controlplane/its1 --for condition=Synced
error: timed out waiting for the condition on controlplanes/its1

```

### Steps To Reproduce

1. Setup a KS installation 
2. Run above commands

### Expected Behavior

The "wait for" should return with condition met

pdettori:
I think I found an explanation - if the generation field and the observedGeneration do not match, it suggests that the controller might not have fully synchronized with the desired state specified in the resource. This discrepancy can cause the kubectl wait command to time out because the system sees this as an indication that syncing isn't complete.
Indeed:
```
$ k -n default wait cps/its1 --for condition=Ready
```
times out. But if I do:
```
$ kubectl patch controlplane its1 --subresource='status' --type=merge -p '{"status": {"observedGeneration": 1}}'
controlplane.tenancy.kflex.kubestellar.org/its1 patched
```
(where "1" is the generation value in the spec)
then it works:
```
$ k -n default wait cps/its1 --for condition=Ready
controlplane.tenancy.kflex.kubestellar.org/its1 condition met
```

PR #325, opened 2025-03-16
🐛 fix: update observedGeneration to match generation in ControlPlane …

asmit27rai:
## Summary

This PR fixes an issue where the `kubectl wait` command times out when waiting for conditions like `Synced` or `Ready` on a `ControlPlane` resource. The root cause is that the `.status.observedGeneration` field is not being updated to match `.metadata.generation`, causing the controller to appear out of sync.

The changes include:
- Updating the `Reconcile` function in `controllers/controlplane_controller.go` to ensure `.status.observedGeneration` is set to `.metadata.generation` when the resource is processed.
- No changes were required for the CRD definition, as the `observedGeneration` field is already correctly defined.

Fixes: #300

### DIFF
--- a/internal/controller/controlplane_controller.go
+++ b/internal/controller/controlplane_controller.go
@@ -115,6 +115,15 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	}
 	hcp := hostedControlPlane.DeepCopy()
 
+	// Update observedGeneration if it doesn't match the current generation
+	if hcp.Status.ObservedGeneration != hcp.Generation {
+		hcp.Status.ObservedGeneration = hcp.Generation
+		if err := r.Status().Update(ctx, hcp); err != nil {
+			log.Error(err, "Failed to update ControlPlane status with observedGeneration")
+			return ctrl.Result{}, err
+		}
+	}
+
 	// finalizer logic
 	if hcp.GetDeletionTimestamp() != nil {
 		if controllerutil.ContainsFinalizer(hcp, kfFinalizer) {


asmit27rai:
@MikeSpreitzer Pls review this PR...

pdettori:
@asmit27rai could you please amend your commit for DCO signing ? See for example [here](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) for instructions.

asmit27rai:
@pdettori Pls review this PR...

internal/controller/controlplane_controller.go
```diff
@@ -115,6 +115,15 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	}
 	hcp := hostedControlPlane.DeepCopy()
 
+	// Update observedGeneration if it doesn't match the current generation
+	if hcp.Status.ObservedGeneration != hcp.Generation {
+		hcp.Status.ObservedGeneration = hcp.Generation
+		if err := r.Status().Update(ctx, hcp); err != nil {
```
pdettori, line 121: There is no need to update the status here - it is updated only in one place which is either UpdateStatusForSyncingSuccess() or UpdateStatusForSyncingError(). The e2e tests are not passing because this update cannot be done before other fields in the status such as the conditions are set. It should be enough to replace the code from line 118 to 126 with just `hcp.Status.ObservedGeneration = hcp.Generation` - could you please verify it works as expected and update the PR? Thanks!

pdettori (changes requested):

internal/controller/controlplane_controller.go
asmit27rai: @pdettori Pls check i updated the PR...

pdettori: thank you @asmit27rai CI tests are now passing. PR looks good.

pdettori (approved):
/lgtm

PR #332, opened 2025-03-30
✨ feature: add new subcommand list-ctx to show all kubeconfig contexts

### LINKED ISSUE #166: feature: Add kflex command to get the list of available contexts
### Feature Description

In some circumstances it would be useful to have a `kflex` command to report the list of available contexts.

### Proposed Solution

Add a new command such as `kflex list-ctx` to return a list of available contexts

### Want to contribute?

### Additional Context

_No response_

tringuyen180303:
Fixes #166
## Summary
I fixed on cmd/ctx/ctx.go file to create a new function to list all the contexts and then also fixed on the cmd/list/main.go to add the function init file. 

Here is the example what will list-ctx look likes:

 ./kflex list-ctx
Available contexts:
  minikube
  wds1
  its1

### DIFF
--- a/cmd/kflex/ctx/ctx.go
+++ b/cmd/kflex/ctx/ctx.go
@@ -58,6 +58,9 @@ func (c *CPCtx) Context(chattyStatus, failIfNone, overwriteExistingCtx, setCurre
 		}
 		fmt.Println(currentContext)
 		return
+	case "list":
+		c.ListContexts()
+		return
 	case "":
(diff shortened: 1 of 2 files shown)


pdettori:
thanks @tringuyen180303 for your PR, but we decided to go with the approach in PR #336 because it list contexts as subcommand of `kflex ctx` which is the main command for dealing with contexts in the kubeflex CLI.

PR #349, opened 2025-05-06
🐛bug fix: Update contributors.md with instructions to fetch upstream tags

### LINKED ISSUE #348: bug: contributors.md needs update
### Describe the bug

Following the steps in Kubefex's docs/contributors.md does not work if contributor is running from a newly created fork without pulling the tags from upstream.

### Steps To Reproduce

Follow steps in `docs/contributors.md` and try to run `kflex into -c` 

### Expected Behavior

should work as described in instructions.

### Additional Context

_No response_

RayyanSeliya:
## Summary
This PR updates the `contributors.md` file to include instructions for fetching tags from the upstream repository. As mentioned in issue #348, without these tags, commands like `kflex init -c` fail because version information is derived from git tags.

The changes:
- Added instructions to set up the upstream remote
- Added instructions to fetch tags from upstream
- Added a note explaining why fetching tags is important

These simple additions should help new contributors avoid the frustration of commands not working after following the guide.

## Related issue(s)
Fixes #348

### DIFF
--- a/docs/contributors.md
+++ b/docs/contributors.md
@@ -13,14 +13,27 @@ Make sure that `${HOME}/go/bin` is in your `$PATH`.
 
 ## How to build kubeflex from source
 
-Clone the repo, build the binaries and add them to your path:
+Clone the repo, set up upstream remote, fetch tags, build the binaries and add them to your path:
 
 ```shell
-git clone https://github.com/kubestellar/kubeflex.git
+# Clone your fork
+git clone https://github.com/<your-username>/kubeflex.git
 cd kubeflex
+
+# Add the upstream repository as a remote
+git remote add upstream https://github.com/kubestellar/kubeflex.git
+
+# Fetch all tags from upstream
+git fetch upstream --tags
+
+# Build the binaries
 make build-all
+
+# Add binaries to your path
 export PATH=$(pwd)/bin:$PATH
 ```
+
+> **Note:** Fetching tags from upstream is important as the version information for KubeFlex binaries is derived from git tags. Without the tags, commands like `kflex init -c` (which initializes KubeFlex and creates a kind cluster) will not work correctly.
 
 ## Setting Up a Testing Cluster for KubeFlex
 
--- a/docs/contributors.md
+++ b/docs/contributors.md
@@ -16,11 +16,11 @@ Make sure that `${HOME}/go/bin` is in your `$PATH`.
 Clone the repo, set up upstream remote, fetch tags, build the binaries and add them to your path:
 
 ```shell
-# Clone your fork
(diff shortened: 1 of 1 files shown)


rxinui:
That looks good to me!

Although, I would be more generic into the clone step. Instead of indicating how to clone the repo, I would specify it with text. Reason being, is the usage of ssh protocol over https, to avoid the hassle of using PAT.

docs/contributors.md
```diff
@@ -13,15 +13,28 @@ Make sure that `${HOME}/go/bin` is in your `$PATH`.
 
 ## How to build kubeflex from source
 
-Clone the repo, build the binaries and add them to your path:
+Clone the repo, set up upstream remote, fetch tags, build the binaries and add them to your path:
 
 ```shell
-git clone https://github.com/kubestellar/kubeflex.git
+# Clone your fork
+git clone https://github.com/<your-username>/kubeflex.git
```
pdettori, line 20: We probably do not need to show the SSH-based git clone (and git remote add) but perhaps mention that in a comment  .. something like
```
# Clone your fork - command only shown for HTTP, adjust if using SSH 
git clone https://github.com/<your-username>/kubeflex.git
```

docs/contributors.md
```diff
@@ -13,15 +13,28 @@ Make sure that `${HOME}/go/bin` is in your `$PATH`.
...
 cd kubeflex
+
+# Add the upstream repository as a remote
+git remote add upstream https://github.com/kubestellar/kubeflex.git
+
```
pdettori, line 25: See comment above - mention the SSH option

pdettori (changes requested):
overall looks good, just two minor comments - thx

pdettori:
> That looks good to me!
> 
> Although, I would be more generic into the clone step. Instead of indicating how to clone the repo, I would specify it with text. Reason being, is the usage of ssh protocol over https, to avoid the hassle of using PAT.

Thanks @rxinui  for helping reviewing - I agree with you that we have to consider also the SSH case, but most commonly one uses https with public GitHub repos, so I suggested a compromise, which is to leave the HTTPS example and write in the comment to adjust the URL if using SSH.

docs/contributors.md
RayyanSeliya: Thanks for the suggestion! I’ll update the file to show only the HTTPS example, with a comment about adjusting the URL for SSH.

docs/contributors.md
RayyanSeliya: I have made the necessary changes !

RayyanSeliya:
@pdettori I have made the suggested changes ! Thankyou for  the feedback

pdettori (approved):
/lgtm - thanks @RayyanSeliya

PR #350, opened 2025-05-06
🐛 bug fix: Retry when PostCreateHook is not found

### LINKED ISSUE #319: bug: No reaction to creation of PostCreateHook
### Describe the bug

At https://github.com/kubestellar/kubeflex/blob/v0.8.1/pkg/reconcilers/shared/postcreate_hook.go#L79 we see that if the PostCreateHook is not found then an error is returned. All the call sites react by putting an error in the ControlPlane's status and claiming successful completion. But looking at the call sites, we see that a re-run later when the PCH exists would produce good results.

### Steps To Reproduce

This is from code inspection.

### Expected Behavior

Eventual consistency.

### Additional Context

_No response_

RayyanSeliya:
## Summary
This PR fixes a bug where creating a control plane would fail if the PostCreateHook doesn't exist yet. 

Changes made:
1. Added a special error type to detect when a hook is missing
2. Modified all reconcilers to check for this error type
3. Added logic to retry after 10 seconds when a hook is not found
4. Keeps the control plane in Available state while waiting for the hook

This allows users to create control planes that reference hooks before creating the hooks themselves.

## Related issue(s)
Fixes #319

### DIFF
--- a/pkg/reconcilers/external/reconciler.go
+++ b/pkg/reconcilers/external/reconciler.go
@@ -18,6 +18,7 @@ package external
 
 import (
 	"context"
+	"time"
 
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/dynamic"
@@ -63,6 +64,11 @@ func (r *ExternalReconciler) Reconcile(ctx context.Context, hcp *tenancyv1alpha1
 	if hcp.Spec.PostCreateHook != nil &&
 		tenancyv1alpha1.HasConditionAvailable(hcp.Status.Conditions) {
 		if err := r.ReconcileUpdatePostCreateHook(ctx, hcp); err != nil {
+			// Check if this is a PostCreateHookNotFoundError
+			if _, ok := err.(*shared.PostCreateHookNotFoundError); ok {
+				// Requeue with a delay to wait for the hook to be created
+				return ctrl.Result{Requeue: true, RequeueAfter: 10 * time.Second}, nil
+			}
 			return r.UpdateStatusForSyncingError(hcp, err)
 		}
 	}
--- a/pkg/reconcilers/host/reconciler.go
+++ b/pkg/reconcilers/host/reconciler.go
@@ -19,6 +19,7 @@ package host
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/dynamic"
@@ -78,6 +79,11 @@ func (r *HostReconciler) Reconcile(ctx context.Context, hcp *tenancyv1alpha1.Con
 	if hcp.Spec.PostCreateHook != nil &&
 		tenancyv1alpha1.HasConditionAvailable(hcp.Status.Conditions) {
 		if err := r.ReconcileUpdatePostCreateHook(ctx, hcp); err != nil {
+			// Check if this is a PostCreateHookNotFoundError
+			if _, ok := err.(*shared.PostCreateHookNotFoundError); ok {
+				// Requeue with a delay to wait for the hook to be created
+				return ctrl.Result{Requeue: true, RequeueAfter: 10 * time.Second}, nil
+			}
 			return r.UpdateStatusForSyncingError(hcp, err)
 		}
 	}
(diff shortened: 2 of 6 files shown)


pdettori:
Thanks @RayyanSeliya for your PR - I think it would be very helpful if you could also add an integration test for this scenario - see  https://github.com/kubestellar/kubeflex/tree/main/test/e2e

rxinui:
Hey! As requested, I'm sharing my input.

From the structure of the multiple `reconciler.go` files, there is a sort of structure pattern on how the error is handle, by the mean of `UpdateStatusForSyncingError` for each and every errors at the reconciler level.

Hence, I would migrate the PostCreateHook error handling inside this https://github.com/kubestellar/kubeflex/blob/ba1a7628782e117d37b4208e7734db1d2209481e/pkg/reconcilers/shared/reconciler.go#L61

In my opinion, this would reduce the changes from 6 files changes to 1.

Also, along the line of being comform, the error handling method seems strange in Go (looks a lot like OOP way of handling errors). I went across the code base and `PostCreateHookNotFoundError` is the only struct for errors. Also the if-statement using type convertion is a bit flaky, might have side effect.

Perhaps, it would be better to:

1. Use the idiomatic go way to define errors using `fmt.Errorf` or `errors.New`
2. Use `errors.As` to compare (or `errors.Is`)

I would recommend to take inspiration on https://pkg.go.dev/io/fs#pkg-constants

RayyanSeliya:
> Thanks @RayyanSeliya for your PR - I think it would be very helpful if you could also add an integration test for this scenario - see https://github.com/kubestellar/kubeflex/tree/main/test/e2e

Thanks for the feedback! I'll add an integration test for this scenario as suggested

RayyanSeliya:
> Hey! As requested, I'm sharing my input.
> 
> From the structure of the multiple `reconciler.go` files, there is a sort of structure pattern on how the error is handle, by the mean of `UpdateStatusForSyncingError` for each and every errors at the reconciler level.
> 
> Hence, I would migrate the PostCreateHook error handling inside this
> 
> https://github.com/kubestellar/kubeflex/blob/ba1a7628782e117d37b4208e7734db1d2209481e/pkg/reconcilers/shared/reconciler.go#L61
> 
> In my opinion, this would reduce from 6 files changes to 1.
> 
> Also, along the line of comformity, the error handling method you implemented seems strange in Go (looks a lot like OOP way of handling errors). I went across the code base and `PostCreateHookNotFoundError` is the only struct for errors. Also the if-statement using type convertion is a bit flaky, might have side effect.
> 
> Perhaps, it would be better to:
> 
> 1. Use the idiomatic go way to define errors using `fmt.Errorf` or `errors.New`
> 2. Use `errors.As` to compare (or `errors.Is`)
> 
> I would recommend to take inspiration on https://pkg.go.dev/io/fs#pkg-constants

Thank you for your helpful feedback! I understand your points about moving the error handling into `UpdateStatusForSyncingError` in the shared reconciler to reduce code changes, and about using Go’s standard error handling methods instead of custom structs.
I will update my code to follow these suggestions and use `fmt.Errorf` or `errors.New` , and compare errors with `errors.As` or `errors.Is` .
Thanks again for your advice and for pointing me to the Go documentation!

rxinui:
Hey, I think it is better to wait for @pdettori approval before applying my recommendation. Sorry I did not mention that! @RayyanSeliya

rxinui:
In general, I will submit my suggestion and vision on something but the decision must be taken from @pdettori or @MikeSpreitzer (not me)

RayyanSeliya:
> Hey, I think it is better to wait for @pdettori approval before applying my recommendation. Sorry I did not mention that! @RayyanSeliya

No worries I completely understand !!

RayyanSeliya:
> In general, I will submit my suggestion and vision on something but the decision must be taken from @pdettori or @MikeSpreitzer (not me)

I agree with you

pdettori:
thanks @rxinui for your review, definitely agree that consolidate the error handling logic in one place (`UpdateStatusForSyncingError`).

MikeSpreitzer:
@RayyanSeliya : do not merge from `main` into your PR's branch.

See https://github.com/kubestellar/kubestellar/blob/194184d6f43edc82945e28d335e9661592a0559d/docs/content/contribution-guidelines/contributing-inc.md#pull-requests .

See https://www.youtube.com/watch?v=tZrNC4pkfyc&t=367s for some relevant clues about using Git.

PR #334, opened 2025-04-02
✨ feature: Add support for alternative kubeconfig naming in ctx command

asmit27rai:
## Summary

This PR adds support for custom naming in kubeconfig entries. Users can now specify a custom name prefix for the cluster, user, and context entries using the `--custom-name` flag in the `kflex ctx` command.

## Related issue

Fixes #307

Issue #308, opened 2025-01-07
bug: improve cobra command code to set subcommand options in a struct that contained the cobra.Command. 


pdettori:
### Describe the bug

There is unnecessary copying of parameters used by the KubeFlex CLI. This can be avoided by defining  the subcommand options  in a struct that contained the cobra.Command rather than (roughly) the other way around. See https://github.com/kubestellar/kubeflex/pull/306#pullrequestreview-2521139829.

### Steps To Reproduce

Look at the code in `cmd/kflex/main.go`

### Expected Behavior

should avoid unnecessary copying of parameters.

rxinui:
After a community meeting, I've shown my intention to help on kubeflex ([also here](https://kubernetes.slack.com/archives/C058SUSL5AA/p1745532439730009?thread_ts=1745349821.923599&cid=C058SUSL5AA)). I'll go ahead and look at this one 👍

PR #351, opened 2025-05-07
✨ add `kflex ctx rename` sub-command

### LINKED ISSUE #307: feature: enable me to work with more than one system
### Feature Description

I switch between multiple distributed systems. One is a triple of OCP clusters on the IBM Cloud. Another is my work testing KubeStellar PRs, which is done with triples of Kind clusters or k3s clusters. I would like to be able to set up a set of kubeconfig contexts for each system, coexisting concurrently. `kflex ctx $cp_name` is not quite up to the task, because it names several sorts of entries in the kubeconfig based on the `$cp_name`: a context, a cluster, and a user. `kubectl config rename-context` is a thing, but renaming a cluster and a user is not. I could do the edits by hand, but that is tedious and easy to mess up.

### Proposed Solution

Add an option on `kflex ctx $cp_name` that conveys an alternative to `$cp_name` to use in the names of the cluster, user, and context.

### Want to contribute?

### Additional Context

_No response_

rxinui:
## Summary

> Add an option on kflex ctx $cp_name that conveys an alternative to $cp_name to use in the names of the cluster, user, and context.

Based on the description given by @MikeSpreitzer, an option `--alias` has been added to `kflex ctx`.

From my understanding, the feature _should_ create an alias entry for a given cp_name to avoid conflict between kubeconfig.

I would like more input from the author, to ensure the need is answered.

### Execution trace

Given the kubeconfig (with its respective cluster and user)

```yaml
- context:
    cluster: test1-cluster
    namespace: default
    user: test1-admin
  name: test1
```

Running `kflex ctx test1 --alias test2` will **replace** the above context,cluster and user to

```yaml
- context:
    cluster: test2-cluster
    user: test2-admin
  name: test2
```

and switch to context `test2`. **Note:** `test2` context is exactly `text1`.

If `kflex ctx test1` is run, as the code is at the moment, `loadAndMergeFromServer` is called which retrieve from the kubeapiserver `test1` context and append it:

```yaml
- context:
    cluster: test2-cluster
    user: test2-admin
  name: test2
- context: # Append
    cluster: test1-cluster
    namespace: default
    user: test1-admin
  name: test1
```

### Strategy

To avoid bringing a lot complexity due to the refactoring issue, I have leverage the current code structure by minimizing changes across files, compared to the following proposition https://github.com/kubestellar/kubeflex/pull/334/files

The logic of --alias is inserted in `cmd/kflex/ctx/ctx.go`. You could notice that some part of the code are sub-code extracted from existent utils functions such as `adjustConfigKeys` with the use of `RenameKey`. The reason why I haven't make use of `adjustConfigKeys` and other existent utils methods is because of their business logic. It seems like to adjust config according to a CP type. This could be refactored and more generic but I would suggest it to be a follow up PR.

### Alternative

1. It could have been possible to leverage the `overwriteExistingCtx` boolean, to override a context within KubeConfig. However, few instructions would be repeated (although idempotent) which are unnecessary.

2. Another alternative would have been to add a new subcommand `kflex ctx rename ${alias}` to first rename (with or without switching context). The advantage is a clearer semantic, as you might edit the context,user,cluster name after a `kflex create ${cp_name}` without switching to it. This could be a follow up PR as well.

## Related issue(s)

Fixes #307


pdettori:
thank you @rxinui

Hey  @MikeSpreitzer - does this approach works for your scenario?

pdettori:
@rxinui could you pls. amend your commit to add DCO sign-off ? The DCO check is not passing - https://github.com/kubestellar/kubeflex/actions/runs/14878929455/job/41801260687?pr=351

pdettori:
I have a general comment on the UX - wondering if `--alias` could be confusing - the result of the command is not an additional alias for the same context (say alias test2 in addition to the existing test1 ). Wondering if instead would be clearer to have something like - using a `rename` subcommand for ctx. 

`kflex ctx rename test1 test2`

rxinui:
@pdettori regarding the `kflex ctx rename test1 test2` refers to the alternative 2. 

I personally think the semantic should be a subcommand `rename` of `kflex ctx`. Yet, as the codebase is at the moment, I wouldn't implement it like because it would go under this 

https://github.com/kubestellar/kubeflex/blob/564fd697edec977fc3d985c91a1cfcfa4925a708/cmd/kflex/ctx/ctx.go#L61-L63

which will be a specific `case` and the issue #308 implementation would completly change the structure.

One of the idea would be to have both at some point with a slight difference
1. `kflex ctx test1 --rename test2` which renames and then switch context to the renamed context
2. `kflex ctx rename test1 test2` which renames the context without switching the current context

Could that be a good direction?

rxinui:
> @rxinui thanks for the clarifications. Wondering if it makes sense to introduce a UX now that we know it has to change later. I would prefer to keep this PR on the back burner, prioritize #308 and after that is completed revisit this PR for the desired UX - I think we both agree that `semantic should be a subcommand rename of kflex ctx.` :)

Hi @pdettori ! Now that a huge refactoring has been made through #352 I have adapted this PR to incorporate @MikeSpreitzer request as `kflex ctx rename` command.

Now the codeflow is becoming more and more smooth 🚀

pdettori (changes requested):
Thanks @rxinui the UX looks much cleaner with `rename` and the code definitely benefit from the refactoring. 

Could you pls. add something about this new feature in the user doc ?

Also, it would be much appreciated to have some test with this - either as unit test or in the e2e tests, to at least try to cover cases such as 
- Renaming a non-existent context.
- Switching to a context after renaming it.
- Handling errors during context deletion or renaming.

cmd/kflex/ctx/rename.go
```diff
@@ -0,0 +1,87 @@
...
+		kconf.CurrentContext = newCtxName
+	} else {
+		fmt.Fprintf(os.Stdout, "switching to hosting cluster context\n")
+		kubeconfig.SwitchToHostingClusterContext(kconf, false)
+	}
+	if err = kubeconfig.WriteKubeconfig(cp.Kubeconfig, kconf); err != nil {
+		return fmt.Errorf("error writing kubeconfig: %v", err)
+	}
+	return nil
+}
```
rxinui, line 87: @pdettori the rename function has been updated to comply with the tests you requested. Now it works as follow:
- if the ctx to rename does not exist, an error is thrown
- if `--switch` is not specified, the current context *does switch* to the hosting cluster context (otherwise it is a config drift)
- ExecuteCtxRename does expect `ctxName` and `newCtxName` as `cp.Name` does not necessarily refers to `ctxName` now that we grant the end-user the ability to name their context.

cmd/kflex/ctx/rename_test.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+	kconf.CurrentContext, err = kubeconfig.GetHostingClusterContext(kconf)
+	if err != nil {
+		return err
+	}
+	if err = kubeconfig.WriteKubeconfig(kubeconfigPath, kconf); err != nil {
+		return fmt.Errorf("error writing kubeconfig: %v", err)
+	}
+	return nil
+}
```
rxinui, line 63: @pdettori Because CLI command mutate the state of kubeconfig, some manual manipulation is required on the kubeconfig.

pkg/kubeconfig/config-util.go
```diff
@@ -76,38 +76,43 @@ func DeleteContext(config *clientcmdapi.Config, cpName string) error {
...
 }
 
-func SwitchToHostingClusterContext(config *clientcmdapi.Config, removeExtension bool) error {
-	if !IsHostingClusterContextPreferenceSet(config) {
-		return fmt.Errorf("hosting cluster preference context not set")
-	}
-
-	// found that the only way to unmarshal the runtime.Object into a ConfigMap
-	// was to use the unMarshallCM() function based on json marshal/unmarshal
+func GetHostingClusterContext(config *clientcmdapi.Config) (string, error) {
```
rxinui, line 84: @pdettori `GetHostingClusterContext` is a new function to return hosting cluster context name. It is handy on the test side and might be in the codebase (probably during the refactoring phase).

pkg/kubeconfig/kubeconfig.go
```diff
@@ -101,13 +102,19 @@ func loadControlPlaneKubeconfig(ctx context.Context, client kubernetes.Clientset
...
+	if kubeconfig == "" {
+		kubeconfig = clientcmd.NewDefaultPathOptions().GetDefaultFilename()
+	}
 	return clientcmd.LoadFromFile(kubeconfig)
 }
 
-func WriteKubeconfig(ctx context.Context, config *clientcmdapi.Config) error {
-	kubeconfig := clientcmd.NewDefaultPathOptions().GetDefaultFilename()
+// Write config into provided kubeconfig file
+func WriteKubeconfig(kubeconfig string, config *clientcmdapi.Config) error {
```
rxinui, line 114: @pdettori before, `WriteKubeconfig` expected `ctx context.Context` as a first parameter, yet it was not used at all within the code. Moreover, this does not make sense as the function name suggests, it *writes the kubeconfig*.

Therefore, it is coherent to replace it with `kubeconfig string` parameter. This enable custom kubeconfig file provided by `--kubeconfig`. Until now, even tho `--kubeconfig` was available for the end-user, it was always writting into `$HOME/.kube/config`. This bug is now partly fixed. In fact, `WriteKubeconfig` is called with an empty string `WriteKubeconfig("")` in many places in the codebase (see line 64 of this file)

This should be fixed in a follow up Issue/PR which I will create and after this one. Are you ok with that?

rxinui (commented):
@pdettori the tests you have requested are now implemented as unit test.

The current state of `kflex` does not allow us to write E2E test (we miss `kflex ctx delete`)

pkg/kubeconfig/kubeconfig.go
pdettori: @rxinui  yes, that works, thank you

pdettori (commented):
@rxinui thanks for adding the unit test, it looks good, would you mind documenting this new command in `docs/users.md` ?

rxinui:
> @rxinui thanks for adding the unit test, it looks good, would you mind documenting this new command in `docs/users.md` ?

@pdettori the documentation has been added and will be maintained when a new `ctx` command will be added, from now on.

If this is /lgtm, I think it is ready to be merged :)

cmd/kflex/ctx/rename_test.go
```diff
@@ -0,0 +1,143 @@
...
+		Cluster:  certs.GenerateClusterName(ctxName),
+		AuthInfo: certs.GenerateAuthInfoAdminName(ctxName),
+	}
+	kconf.Clusters[certs.GenerateClusterName(ctxName)] = api.NewCluster()
+	kconf.AuthInfos[certs.GenerateAuthInfoAdminName(ctxName)] = api.NewAuthInfo()
+	kconf.CurrentContext = hostingClusterContextMock
+	kubeconfig.SetHostingClusterContextPreference(kconf, nil)
+	if err := kubeconfig.WriteKubeconfig(kubeconfigPath, kconf); err != nil {
+		return fmt.Errorf("error writing kubeconfig: %v", err)
+	}
```
rxinui, line 45: @pdettori to pass the test, instead of relying on a real kubeconfig file, i created a partial one which will be deleted on teardown.

pdettori (approved):
/lgtm

PR #352, opened 2025-05-08
🐛 refactoring and design change on the usage of spf13/cobra

### LINKED ISSUE #308: bug: improve cobra command code to set subcommand options in a struct that contained the cobra.Command.

rxinui:
# Proof-of-concept - READY TO MERGE

## Summary

Before this PR, a couple of parameters, types and others have been copy paste and did not leverage the power that spf13/cobra offers (reusability, modularity...)

To solve this problem, I decided to change the codebase design at the `cmd/kflex/` level. 

For this proof of concept, **ONLY** the commands listed below have been refactored and the code as it is currently would not build.
- `kflex create`
- `kflex list`

Overall, the design change vows to:
- simplify the codeflow which will make the OSS adoption by contributors easier
- avoid side-effects as much as possible, which could explain the number of bugs reported
- improve readability of the codebase with better naming convention, remove unecessary struct/types, better semantic
- be more idiomatic. My references are [Learn Go the Idiomatic way](https://www.oreilly.com/library/view/learning-go-2nd/9781098139285/) and [100 Mistakes in Go and how to avoid it](https://100go.co/)

**Note:** two types of comments are added within the codebase for this PoC
- `// REFACTOR:` explain the refactor made and consequences
- `// REFACTOR?` indicates a question -- should it be refactor?

## Code design

The new code design aims to alleviate the hassle of defining everything within the root file and aims to isolate command code within their own package to limit the dependencies between root and its commands.

### Project structure

It is inspired by domain-driven dev. To avoid cyclic import, the structure policy is the following:

```
cmd/kflex
├── common                   # shared package -- only package which does not represent commands but reusable types, functions and flags
│   ├── cp.go                # shared 
│   └── flags.go             # shared flags -- define global flags constants
├── create
│   └── create.go
├── ctx                      # package -- indicates command
│   ├── ctx.go               # command -- noticeable because matches package name
│   └── list.go              # any other files indicates they are subcommands
├── list
│   └── list.go
...
└── main.go                  # root command -- where kflex is defined
```

1. **Top-to-bottom direction**. A command cannot import/use root resources nor resources from its siblings `root --> commands --> subcommands ...`
2. **Shared package**. Only `common` package stores reusable resources callable by root command, its commands and subcommands. `common` is the only package not defining a command.
3. **Packages represent commands**. Definition of a command and its subcommand are made within its package: flags definition, code logic, subcommands.
4. **Files within command package**. A command is named after its package name, hence any file within a package with different naming would represent a subcommands. For instance `ctx/ctx.go` defines the `kflex ctx` whereas `ctx/list.go` defines the subcommand of ctx `kflex ctx list`. (NOT IMPLEMENTED IN THIS PoC !!!) --> *Note: in relation with #351, this structure will then allow to easily implement `kflex ctx rename` by creating `ctx/rename.go` file*

Proceding as follow makes every commands business-independant and has separation of concerns.

### spf13/cobra specificities

To avoid global variables within the root file, we make use of `FlagSet` which does not store CLI value within these variables. Instead, each command will define 

1. `func Command() *cobra.Command` which returns the command with its flags and execution code initialised.
2. `func execute(arguments...)` which represent the logic of the command (ie. create controlplane)

This guarentee that flags values passed cannot be modified by another process, which was a possibility before by the mean of global variables to store flag values. Now, `command.Flags().StringP` replaces `command.Flags().StringVarP`.

**NOTE:** the only global variables that may remains are those required by `gorelease` to inject values at compile time (such as `kflex version`).

To avoid memory consumption, `command.Flags()` is no longer call for each options. Instead, it instantiated once for one command in one execution, as shown here 
https://github.com/kubestellar/kubeflex/compare/main...rxinui:kubeflex:bug/308?expand=1#diff-20056e0c2eba9525b066cf3155bfc265204830faa73252b3ea4be36f4dd3b736R56-R65

Unnecessary `if-test` are replaced by the mean of cobra default values. See https://github.com/kubestellar/kubeflex/compare/main...rxinui:kubeflex:bug/308?expand=1#diff-20056e0c2eba9525b066cf3155bfc265204830faa73252b3ea4be36f4dd3b736R69-R73

### Better readability

Now, flags name are moved as constants. It will make OSS contribution easier when existing option for a given command should also be added to another one (for instance, `--set` flag define in `create` should be added to `ctx`). Shared flags are define in `common` package while command-specific flag are define within its command definition.

In the legacy codebase, a tricky/bad usage of variable name is done with control plane. In fact, `cp` is currently used as `var cp CP` but it gets overshadow by `tenancyv1alpha1.ControlPlane` type. This has been fix by improving variable naming. Now `cp` strictly refers to `common.CP` whereas `controlPlane` refers to `tenancyv1alpha1.ControlPlane`. See https://github.com/kubestellar/kubeflex/compare/main...rxinui:kubeflex:bug/308?expand=1#diff-97c377915d8ccea8cbc135426b489f90bae558e022c7c02310f2a02ecca99583R53-R89

(a) All struct variants of `CP` such as `CPCreate` or `CPList` are deleted as they are confusing and do not bring values. Instead, `CP` is now moved to `common` package and functions has been adapted to receive `CP` as function parameter rather than method parameter. See https://github.com/kubestellar/kubeflex/compare/main...rxinui:kubeflex:bug/308?expand=1#diff-97c377915d8ccea8cbc135426b489f90bae558e022c7c02310f2a02ecca99583R41

(b) Moreover, `CP` is now meant to be instantiated by the use of `NewCP` function. This factory implements the Option design pattern which allows optional parameters. That simplify a lot the codebase. We move from the old way

```go
cp := del.CPDelete{
			CP: common.CP{
				Ctx:        createContext(),
				Name:       args[0],
				Kubeconfig: kubeconfig,
			},
		}
```

to simply

```go
cp := common.NewCP(kubeconfig, common.WithName(args[0]))
```

In fact, `del.CPDelete` becomes irrelevant because of (a) and `NewCP` (b) embed `createContext()` process. With the option design pattern, it makes it easier with command that does not expect a `CP.Name` such as `listCtxCommand`. In such cases, we obtain

```go
cp := common.NewCP(kubeconfig)
```

## Related issue(s)

Fixes #308

### DIFF
--- a/cmd/kflex/adopt/adopt.go
+++ b/cmd/kflex/adopt/adopt.go
@@ -77,14 +77,14 @@ func Command() *cobra.Command {
 			adoptedContext, _ := flagset.GetString(AdoptedContextFlag)
 			adoptedURLOverride, _ := flagset.GetString(URLOverrideFlag)
 			adoptedTokenExpirationSeconds, _ := flagset.GetInt64(ExpirationSecondsFlag)
-			cp := CPAdopt{
+			cpAdopt := CPAdopt{
 				CP:                            common.NewCP(kubeconfig, common.WithName(args[0])),
 				AdoptedKubeconfig:             adoptedKubeconfig,
 				AdoptedContext:                adoptedContext,
 				AdoptedURLOverride:            adoptedURLOverride,
 				AdoptedTokenExpirationSeconds: adoptedTokenExpirationSeconds,
 			}
-			ExecuteAdopt(cp, postCreateHook, hookVars, chattyStatus)
+			ExecuteAdopt(cpAdopt, postCreateHook, hookVars, chattyStatus)
 		},
 	}
 	flagset := command.Flags()
@@ -99,33 +99,33 @@ func Command() *cobra.Command {
 }
 
 // Adopt a control plane from another cluster
-func ExecuteAdopt(cp CPAdopt, hook string, hookVars []string, chattyStatus bool) {
+func ExecuteAdopt(cpAdopt CPAdopt, hook string, hookVars []string, chattyStatus bool) {
 	done := make(chan bool)
 	var wg sync.WaitGroup
 	cx := cont.CPCtx{}
 	cx.ExecuteCtx(chattyStatus, false, false, false)
 
 	controlPlaneType := tenancyv1alpha1.ControlPlaneTypeExternal
-	util.PrintStatus(fmt.Sprintf("Adopting control plane %s of type %s ...", cp.Name, controlPlaneType), done, &wg, chattyStatus)
+	util.PrintStatus(fmt.Sprintf("Adopting control plane %s of type %s ...", cpAdopt.Name, controlPlaneType), done, &wg, chattyStatus)
 
-	cl, err := kfclient.GetClient(cp.Kubeconfig)
+	cl, err := kfclient.GetClient(cpAdopt.Kubeconfig)
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Error getting kubeflex client: %v\n", err)
 		os.Exit(1)
 	}
 
-	clientsetp, err := kfclient.GetClientSet(cp.Kubeconfig)
+	clientsetp, err := kfclient.GetClientSet(cpAdopt.Kubeconfig)
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Error getting clientset: %v\n", err)
 		os.Exit(1)
 	}
 
-	if err := applyAdoptedBootstrapSecret(clientsetp, cp.Name, getBootstrapKubeconfig(cp), cp.AdoptedContext, cp.AdoptedURLOverride); err != nil {
+	if err := applyAdoptedBootstrapSecret(clientsetp, cpAdopt.Name, getBootstrapKubeconfig(cpAdopt), cpAdopt.AdoptedContext, cpAdopt.AdoptedURLOverride); err != nil {
 		fmt.Fprintf(os.Stderr, "error creating adopted cluster kubeconfig: %v\n", err)
 		os.Exit(1)
 	}
-	// REFACTOR? is cp being pointer of (*CPAdopt) make sense?
-	controlPlane, err := common.GenerateControlPlane(cp.Name, string(controlPlaneType), "", hook, hookVars, &cp.AdoptedTokenExpirationSeconds)
+	// REFACTOR? is cpAdopt being pointer of (*CPAdopt) make sense?
+	controlPlane, err := common.GenerateControlPlane(cpAdopt.Name, string(controlPlaneType), "", hook, hookVars, &cpAdopt.AdoptedTokenExpirationSeconds)
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "error generating control plane object: %v\n", err)
 		os.Exit(1)
@@ -277,13 +277,12 @@ func isValidServerURL(serverURL string) error {
 	return nil
 }
 
-// REFACTOR: no reason to be a pointer
-func getBootstrapKubeconfig(cp CPAdopt) string {
-	if cp.AdoptedKubeconfig != "" {
-		return cp.AdoptedKubeconfig
+func getBootstrapKubeconfig(cpAdopt CPAdopt) string {
+	if cpAdopt.AdoptedKubeconfig != "" {
+		return cpAdopt.AdoptedKubeconfig
 	}
-	if cp.Kubeconfig != "" {
-		return cp.Kubeconfig
+	if cpAdopt.Kubeconfig != "" {
+		return cpAdopt.Kubeconfig
 	}
 	return getKubeConfigFromEnv()
 }
--- a/cmd/kflex/adopt/adopt.go
+++ b/cmd/kflex/adopt/adopt.go
@@ -84,7 +84,7 @@ func Command() *cobra.Command {
 				AdoptedURLOverride:            adoptedURLOverride,
 				AdoptedTokenExpirationSeconds: adoptedTokenExpirationSeconds,
 			}
-			execute(cp, postCreateHook, hookVars, chattyStatus)
+			ExecuteAdopt(cp, postCreateHook, hookVars, chattyStatus)
 		},
 	}
 	flagset := command.Flags()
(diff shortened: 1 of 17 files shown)


rxinui:
@pdettori as discussed, I'm raising an intermediate PR as a Proof-of-concept before implementing the whole refactoring process, so we can decide together what is right and wrong.

CC: @clubanderson

rxinui:
**NOTE**: this branch has been checkout from #351 hence the changes in `pkg/kubeconfig.go`. Please ignore that.

cmd/kflex/common/cp.go
```diff
@@ -31,10 +34,44 @@ type CP struct {
...
+}
+
+// Create a new system context with logger
+func createContext() context.Context {
+	zapLogger, _ := zap.NewDevelopment(zap.AddCaller())
+	logger := zapr.NewLoggerWithOptions(zapLogger)
+	return logr.NewContext(context.Background(), logger)
+}
+
+// REFACTOR: the usage of the variable `cp` as *common.CP and then overshadowed by `cp` as tenancyv1alpha1.ControlPlane was tricky. Took a while to understand. Hence, `cp` from tenancyv1alpha1 becomes `controlPlane` and common.CP is referred by `cp`
```
pdettori, line 72: good point @rxinui - it is much more clear now.

cmd/kflex/common/flags.go
```diff
@@ -0,0 +1,26 @@
...
+package common
+
+import "k8s.io/client-go/tools/clientcmd"
+
+const (
+	KubeconfigFlag   = clientcmd.RecommendedConfigPathFlag
+	ChattyStatusFlag = "chatty-status"
+	VerbosityFlag    = "verbosity"
+	SetFlag          = "set"
+)
```
pdettori, line 26: yes, this was overdue, there was repetition of the flags

cmd/kflex/create/create.go
```diff
@@ -32,61 +33,101 @@ import (
...
+	BKTypeDefault        = string(tenancyv1alpha1.BackendDBTypeShared) // REFACTOR? local to create or common
+	CTypeDefault         = string(tenancyv1alpha1.ControlPlaneTypeK8S) // REFACTOR? local to create or common
+	ControlPlaneTypeFlag = "type"
+	BackendTypeFlag      = "backend-type"
+	PostCreateHookFlag   = "postcreate-hook"
+	ChattyStatusFlag     = "chatty-status"
+)
+
+// REFACTOR: removed variables such as `hookVars` as they are used by multiple commands (create and adopt...). It should be defined locally to each command package instead of pointing to the same variable defined in main avoiding extreme edge cases.
+
```
pdettori, line 47: good catch

pdettori (commented):
Thanks @rxinui - overall it looks good to me - there is more clarity is the usage of some variables and the structure you propose for the create and list command is what I had in mind.

rxinui:
# Refactoring is now complete

kflex commands that works

- create
- ctx
- ctx get
- ctx list
- help
- delete
- list

kflex command with a bug

- version ==> in fact, the version and build date are injected as LDFLAGS from the Makefile. I moved them from package main to package common and it seems that the path given in the makefile is not working. I would like some inputs from you there.

kflex commands not tested

- adopt
- completion
- init

## Notes

- Some potential refactoring can be done but haven't to avoid any big breaking. 
- Another refactoring phase will follow once this phase is confirmed.

cmd/kflex/create/create.go
```diff
@@ -32,61 +33,99 @@ import (
...
-	common.CP
+// defaults
+const (
+	BKTypeDefault        = string(tenancyv1alpha1.BackendDBTypeShared) // REFACTOR? local to create or common
+	CTypeDefault         = string(tenancyv1alpha1.ControlPlaneTypeK8S) // REFACTOR? local to create or common
+	ControlPlaneTypeFlag = "type"
+	BackendTypeFlag      = "backend-type"
+)
+
+// REFACTOR: removed variables such as `hookVars` as they are used by multiple commands (create and adopt...). It should be defined locally to each command package instead of pointing to the same variable defined in main avoiding extreme edge cases.
```
pdettori, line 44: same comment as before - this was already done

cmd/kflex/create/create.go
```diff
@@ -32,61 +33,99 @@ import (
...
+			postCreateHook, _ := flagset.GetString(common.PostCreateHookFlag)
+			hookVars, _ := flagset.GetStringArray(common.SetFlag)
+			cp := common.NewCP(kubeconfig, common.WithName(args[0]))
+			// create passing the control plane type and backend type
+			ExecuteCreate(cp, cpType, backendType, postCreateHook, hookVars, chattyStatus)
+		},
+	}
+
+	flagset := command.Flags()
+	// REFACTOR: putting CTypeDefault as default value, hence if empty string given to the flag, it picks up
```
pdettori, line 68: comment should no longer be necessary

cmd/kflex/create/create.go
```diff
@@ -32,61 +33,99 @@ import (
...
+	flagset.StringP(ControlPlaneTypeFlag, "t", CTypeDefault, "type of control plane: k8s|ocm|vcluster")
+	// REFACTOR: same than CTypeDefault for BKTypeDefault
```
pdettori, line 70: ditto

pdettori (commented):
In addition to inline testing, here are a result of manual testing:
`kflex list` does not print the names of the control planes
```
$ kflex list
Control Planes:
NAME                 SYNCED     READY      TYPE       AGE       
                     True       True       vcluster   18h39m55s 
                     True       True       k8s        18h39m55s 
 ```                     
 e2e testing has passed and has exercised most commands

File: cmd/kflex/adopt/adopt.go [hunk 7]
```diff
@@ -50,42 +60,78 @@ type CPAdopt struct {
...
-	if err := applyAdoptedBootstrapSecret(clientsetp, c.Name, bootstrapKubeconfig, c.AdoptedContext, c.AdoptedURLOverride); err != nil {
+	if err := applyAdoptedBootstrapSecret(clientsetp, cp.Name, getBootstrapKubeconfig(cp), cp.AdoptedContext, cp.AdoptedURLOverride); err != nil {
 		fmt.Fprintf(os.Stderr, "error creating adopted cluster kubeconfig: %v\n", err)
 		os.Exit(1)
 	}
-
-	cp, err := common.GenerateControlPlane(c.Name, string(controlPlaneType), "", hook, hookVars, &c.AdoptedTokenExpirationSeconds)
+	// REFACTOR? is cp being pointer of (*CPAdopt) make sense?
+	controlPlane, err := common.GenerateControlPlane(cp.Name, string(controlPlaneType), "", hook, hookVars, &cp.AdoptedTokenExpirationSeconds)
 	if err != nil {
```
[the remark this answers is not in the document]
rxinui: Agree, on it

rxinui:
@pdettori could you please retry `kflex list` and confirm its behaviour?

pdettori:
@rxinui now list works - thanks

pdettori (approved):
/lgtm - thank you @rxinui !

Issue #356, opened 2025-05-10
bug: kflex delete doesn't work on faulty context


rxinui:
## Description

Currently `kflex delete $myctx` does not work if `$myctx` is faulty.

*Faulty* in my case refers to a context where its connection with apiserver is not working:

```
└─[$] <git:(new/307)> k config use-context test2
Switched to context "test2".
└─[$] <git:(new/307)> k get pods
E0510 15:25:46.097976   76461 memcache.go:265] "Unhandled Error" err="couldn't get current server API group list: Get \"https://test1.localtest.me:9443/api?timeout=32s\": tls: failed to verify certificate: x509: certificate is valid for ingress.local, not test1.localtest.me"
E0510 15:25:46.107259   76461 memcache.go:265] "Unhandled Error" err="couldn't get current server API group list: Get \"https://test1.localtest.me:9443/api?timeout=32s\": tls: failed to verify certificate: x509: certificate is valid for ingress.local, not test1.localtest.me"
```

### Steps To Reproduce

1. Create a dummy context with invalid data named `test1`
2. Test the context using `kflex ctx test1`
3. Verify the context is not able to communicate with apiserver `kubectl get pods`
4. Delete the current faulty context `kflex ctx test1` 

### Expected Behavior

I'd assume that `kflex delete $myctx` should work **regardless of the status of the context**, especially when the context is faulty.

rxinui:
I'll look at it later :)

asmit27rai:
@pdettori @rxinui Can I take this issue?

rxinui:
Hey! I did start something since I reported it, to share the info.

But since the issue isn't assign to me, I do not see any harm of having two PRs on it.

rxinui:
After reflexion, **this isn't a bug but the expected behaviour**.

`kflex delete` is, according to its usage:
> Delete a control plane instance and switches the context back to the hosting cluster context

Therefore, this is not a bug, but the behaviour intended. My bad.

However, this _mistake_ leads to the creation of a new feature: `kflex ctx delete`

As #351 introduce a `kflex ctx rename` it sounds logical to be able to delete contexts using `kflex ctx delete`.

pdettori:
@rxinui you are correct that "this isn't a bug but the expected behaviour." - I agree that `kflex ctx delete` could be a good addition to solve potential orphaned contexts.

PR #357, opened 2025-05-10
 🐛 fix: Allow deletion of contexts with invalid cluster connections

asmit27rai:
## Summary

🐛 Fix kflex delete command for faulty contexts:
- Forces switch to hosting cluster context before operations
- Handles missing control plane resources gracefully
- Ensures kubeconfig context cleanup even with invalid cluster connection
- Improves error handling for kubeconfig modifications

Fixes: #356


pdettori:
This seems a duplicate of #359

PR #359, opened 2025-05-10
✨ add kflex ctx delete command 

### LINKED ISSUE #358: feature: delete ctx (useful when faulty or testing phase)
### Feature Description

See [original discussion](https://github.com/kubestellar/kubeflex/issues/356#issue-3054147808)

> After reflexion, **this isn't a bug but the expected behaviour**.
> 
> `kflex delete` is, according to its usage:
> 
> > Delete a control plane instance and switches the context back to the hosting cluster context
> 
> Therefore, this is not a bug, but the behaviour intended. My bad.
> 
> However, this _mistake_ leads to the creation of a new feature: `kflex ctx delete`
> 
> As [#351](https://github.com/kubestellar/kubeflex/pull/351) introduce a `kflex ctx rename` it sounds logical to be able to delete contexts using `kflex ctx delete`.

When testing or developing with kubeflex, it happens that multiple fake contexts are created. `kflex delete` only interact with control plane. Therefore, if a context is orphan then it is not possible to properly delete it from Kubeconfig and easy to mess up (cf. must delete context, user and cluster).

### Proposed Solution

Add a new subcommand `kflex ctx delete $ctx` to delete $ctx, its associated cluster and user.

### Want to contribute?

### Additional Context

_No response_

rxinui:
## Summary

Proposal PR -- as issue is not yet accepted, the following solution is proposed to give details and working implementation

`kflex ctx delete` subcommand is now added to allow user removing orphans context (created during dev phase).

This command properly clean useless contexts and its cluster and user of the kubeconfig.

**Note:** in the continuity of #352 and its refactoring process, the following has been added to improve the codeflow:

### Continuous improvement 

- **Explicit naming to avoid confusion**. `kubeconfig.DeleteContext` is renamed to `kubeconfig.DeleteAll` as the method delete the triple **context,cluster,user** from the kubeconfig, not just the context as implied by the signature.

- **Better error handling**. Instead of abruptly stop the codeflow using
```
fmt.FPrintf(err) // first display error
os.exit(1) // then exit
```
We replace `cobra.Command.Run` by `cobra.Command.RunE` which returns an `error` at the CLI-level runtime (as [written here](https://github.com/kubestellar/kubeflex/pull/359/files#diff-96eedf9e25c0f4a1893713a906be511d997c1709b08edc1f054ed5001cee7456R35-R40))

Leveraging that, we can simply return an error within our functions which reduce the number of instructions and make the code more idiomatic (reference https://100go.co/#ignoring-when-to-wrap-an-error-49):
```
return fmt.Errorf("error: %v", err) // transformation of new error using %v
```

### Notes

@pdettori if you approve this continuous improvement, the other commands and subcommands will need to be updated in a follow up PR to include this refactoring.

At the moment only `cmd/kflex/ctx/delete.go` includes the changes.

## Related issue(s)

Fixes #358

### DIFF
--- a/cmd/kflex/ctx/delete.go
+++ b/cmd/kflex/ctx/delete.go
@@ -1,0 +1,63 @@
+/*
+Copyright 2023 The KubeStellar Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package ctx
+
+import (
+	"fmt"
+	"sync"
+
+	"github.com/kubestellar/kubeflex/cmd/kflex/common"
+	"github.com/kubestellar/kubeflex/pkg/kubeconfig"
+	"github.com/kubestellar/kubeflex/pkg/util"
+	"github.com/spf13/cobra"
+)
+
+func CommandDelete() *cobra.Command {
+	return &cobra.Command{
+		Use:   "delete CONTEXT",
+		Short: "Delete a context",
+		Long:  `Delete a context in the kubeconfig file`,
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			cmd.SilenceUsage = true
+			kubeconfig, _ := cmd.Flags().GetString(common.KubeconfigFlag)
+			chattyStatus, _ := cmd.Flags().GetBool(common.ChattyStatusFlag)
+			cp := common.NewCP(kubeconfig, common.WithName(args[0]))
+			return ExecuteCtxDelete(cp, chattyStatus)
+		},
+	}
+}
+
+// Execute kflex ctx delete
+func ExecuteCtxDelete(cp common.CP, chattyStatus bool) error {
+	var wg sync.WaitGroup
+	done := make(chan bool)
+	util.PrintStatus("Deleting context", done, &wg, chattyStatus)
+	kconf, err := kubeconfig.LoadKubeconfig(cp.Kubeconfig)
+	if err != nil {
+		return fmt.Errorf("error loading kubeconfig: %v", err)
+	}
+	if err = kubeconfig.DeleteAll(kconf, cp.Name); err != nil {
+		return fmt.Errorf("error deleting context %s from kubeconfig: %v", cp.Name, err)
+	}
+	if err = kubeconfig.WriteKubeconfig(cp.Kubeconfig, kconf); err != nil {
+		return fmt.Errorf("error writing kubeconfig: %v", err)
+	}
+	done <- true
+	wg.Wait()
+	return nil
+}
--- a/cmd/kflex/ctx/delete.go
+++ b/cmd/kflex/ctx/delete.go
@@ -18,11 +18,10 @@ package ctx
 
 import (
 	"fmt"
-	"sync"
+	// "sync"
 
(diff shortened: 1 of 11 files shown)


pdettori:
@rxinui I think this is a good idea - just couple of clarifications to see if I understand correctly:

- if there are two orphan contexts `oc1` and `oc2`  running `kflex ctx delete` would delete `oc1` and `oc2` and associated clusters/users in the context - but it should not touch any other context associated with a control plane on the server.
- If there are no orphan contexts it should not do anything (no op)

Is this correct ?

Also, it would be helpful to show what actions are done if deletes are performed - like "found 2 orphan context - deleting context oc1 etc."

rxinui:
>     * if there are two orphan contexts `oc1` and `oc2`  running `kflex ctx delete` would delete `oc1` and `oc2` and associated clusters/users in the context - but it should not touch any other context associated with a control plane on the server.

Not quite, but I like this idea too. At the moment what has been implemented was a regular `delete {ctx}` command, where the orphan ctx should be specified.

What you suggest is actually a good feature, but then an algorithm must be created to **detect an orphan context**.

>     * If there are no orphan contexts it should not do anything (no op)

To avoid potential user-mistake, I would still require an argument to `kflex ctx delete {ctx}`. To implement the deletion of all orphan contexts as discussed above, i would define a special argument such as `--orphans` to specifically target them.

By doing so, we protect the end-user from mistakenly delete a context. Following that direction, perhaps guards can be added to ask confirmation for deleting a context (especially) if it is not an orphan one.

> Also, it would be helpful to show what actions are done if deletes are performed - like "found 2 orphan context - deleting context oc1 etc."

Yes totally agree!

pdettori:
@rxinui thanks for the clarification, perhaps it was the phrase `allow user removing orphans context ` that threw me off. I think now I understand and I agree with the current direction. I think we can leave the problem of detection of orphan contexts to a different issue/PR. For this one, I think it's a good addition, I would just recommend to add some documentation in `users.md`. Would be also good to exercise this in some tests.

rxinui:
@pdettori I have updated `docs/users.md` and added unit+e2e test to `kflex ctx delete`. Everything is 🟢

cmd/kflex/ctx/delete.go
```diff
@@ -0,0 +1,67 @@
...
+			cp := common.NewCP(kubeconfig)
+			return ExecuteCtxDelete(cp, args[0], chattyStatus)
+		},
+	}
+}
+
+// Execute kflex ctx delete
+func ExecuteCtxDelete(cp common.CP, ctxName string, chattyStatus bool) error {
+	var wg sync.WaitGroup
+	done := make(chan bool)
```
pdettori, line 48: consider adding `defer wg.Done() ` to ensure `done` channel is closed in case of errors

cmd/kflex/ctx/delete_test.go
```diff
@@ -0,0 +1,81 @@
...
+	clusterName := certs.GenerateClusterName(ctxName)
+	authInfoName := certs.GenerateAuthInfoAdminName(ctxName)
+	if _, ok := kconf.Contexts[ctxName]; !ok {
+		t.Errorf("context '%s' still present in kubeconfig", ctxName)
+	}
+	if _, ok := kconf.Clusters[clusterName]; !ok {
+		t.Errorf("cluster '%s' still present in kubeconfig", clusterName)
+	}
+	if _, ok := kconf.AuthInfos[authInfoName]; !ok {
+		t.Errorf("user '%s' still present in kubeconfig", authInfoName)
```
pdettori, line 53: the messages seem wrong - this is checking that user, context etc. is NOT present, no?

pdettori (commented):
overall looks good, just couple of minor points to address

cmd/kflex/ctx/delete.go
rxinui: will do

pdettori:
@rxinui - I think we can accept this proposal and change the title to remove proposal - could you also open an associated issue for this ?

rxinui:
@pdettori yes will do. Right now i'm reviewing your comment. It turns out the print message is correct, what is not is the if-statement. Working on it.

PR #367, opened 2025-05-16
✨  Update the real source of the kube-rbac-proxy image ref

MikeSpreitzer:
## Summary
This PR updates the reference to the kube-rbac-proxy container image, in the real source of the reference.

## Related issue(s)

Fixes #

### DIFF
--- a/config/default/manager_auth_proxy_patch.yaml
+++ b/config/default/manager_auth_proxy_patch.yaml
@@ -41,7 +41,7 @@ spec:
           capabilities:
             drop:
               - "ALL"
-        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.13.1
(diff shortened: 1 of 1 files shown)


MikeSpreitzer:
Now I see this:

```console
$ find * -type f -exec grep /kube-rbac-proxy \{\} \; -print -exec echo \;
        image: quay.io/brancz/kube-rbac-proxy:v0.18.0
chart/templates/operator.yaml

# the auth proxy (https://github.com/brancz/kube-rbac-proxy)
config/rbac/kustomization.yaml

        image: quay.io/brancz/kube-rbac-proxy:v0.18.0
config/default/manager_auth_proxy_patch.yaml
```

pdettori (approved):
/lgtm

PR #373, opened 2025-05-19
🐛 add controlplane metadata at creation

### LINKED ISSUE #372: bug: kflex ctx rename breaks controlplane traceability
### Describe the bug

When running rename context command, it is impossible to determine which context is associated to a kubeflex controlplane.

### Steps To Reproduce

```bash
kflex create mycp # creates a controlplane and context `mycp`
kflex ctx rename mycp newcp # rename mycp's context to `newcp`
```

The controlplane `mycp` is now accessible through context `newcp`. However, there is no trace stating that `newcp` is linked to `mycp` controlplane.

Therefore, after renaming a context, it becomes impossible to determine a whether a context is associated to a kubeflex controlplane. For instance, running the following will create an error

```bash
kflex delete newcp 
# error because `delete` expect a controlplane (mycp) and not its context (newcp)
```

### Expected Behavior

Have a trace indicating the association kubeflex controlplane-context within a user kubeconfig.

### Additional Context

In a follow up issue, create an algorithm which detects a controlplane from a context name. This algorithm will be used to run `kflex delete` on a context name and retrieve the associated controlplane.

Hence, `kflex delete newcp` will be converted to `kflex delete mycp` and result as a success instead of an error. (NOTE: it will be smart to add an option, to prevent mistake deletion such as `--resolve-cp`)

rxinui:
## Summary

When a control plane is created using `kflex create $cp`, user kubeconfig file is updated with kubeflex metadata leveraging [kubeconfig.v1/NamedExtension](https://kubernetes.io/docs/reference/config-api/kubeconfig.v1/#NamedExtension)

An extension is created locally to the associated context stating the controlplane it is linked to. As extension represents metadata for tooling, this information will be used to resolve a controlplane name through a context.

Instead of having:
```yaml
- context:
    cluster: kflex-cp1-cluster
    user: kflex-cp1-admin
  name: kflex-cp1
```

kubeconfig will look like:
```yaml
- context:
    cluster: kflex-cp1-cluster
    extensions:
    - extension:
        data:
          kflex-controlplane-name: kflex-cp1    # associated controlplane
          kflex-initial-ctx-name: kflex-cp1     # first context associated
        metadata:
          creationTimestamp: null
      name: kubeflex
    user: kflex-cp1-admin
  name: kflex-cp1
```

This way, running `kflex ctx rename kflex-cp1 cp2` will  still allow a user to identify its controlplane through its context.

In a followup PR, this extension will be used to:
- `kflex delete` a controlplane using its context value
- `kflex ctx list` display which context is managed by kubeflex and its controlplane

Also, this extension could serve as **preventive guard**.

In fact, it could be problematic to let a user delete a context that is not managed by kubeflex (say openshift production context/cluster/user). Therefore, by looking up for the extension, if not present, kflex can throw a warning to prevent the unintentional deletion of a context/cluster/user.

## Related issue(s)

Fixes #372


pdettori:
@rxinui - I have added the OWNERS file

rxinui:
@pdettori great!

Now with the prow installed, i guess /lgtm and /approve are required step to the process

pdettori:
@rxinui some comment on the use of the extension - currently we use the extension to "tag" the context associated with the hosting cluster. This way we can always do `kflex ctx` to switch context back to the hosting cluster context. 

Example:

```yaml
preferences:
  extensions:
  - extension:
      data:
        kflex-initial-ctx-name: kind-kubeflex
      metadata:
        creationTimestamp: null
        name: kflex-config-extension-name
    name: kflex-config-extension-name
 ```
 
  That `kflex-initial-ctx-name` is not supposed to change to `kflex-cp1` after we create cp1 or another control plane.
  
  I think we can use this extension mechanism, but should not change the value of `kflex-initial-ctx-name` as that is determined at the time of initial setup of kubeflex in the hosting cluster.
  
  Does the extension schema allow to add structured data ? E.g.
  
 ```yaml
preferences:
  extensions:
  - extension:
      data:
        kflex-initial-ctx-name: kind-kubeflex
        control-planes:
         - name: cp1
            contextName: cp1
         - name: cp2
            contextName: cp2-renamed
      metadata:
        creationTimestamp: null
        name: kflex-config-extension-name
    name: kflex-config-extension-name
 ```

rxinui:
There are 3 scopes of extensions usable in our cases, in kubeconfig.

1. Global under `extensions:`
2. Context scoped `contexts[*].extensions`
3. Preference scoped (meant to alter kubectl behaviour) `preferences.extensions`

To assign a controlplane to a context, I've chosen 2.

What is done for the current switch to hosting cluster, 3 has been chosen.

pdettori:
@rxinui thank you for the clarification. I am fine with context scoped for keeping the mapping between control plane name and context name. As for the `kflex-initial-ctx-name` I don't think we should repeat that for each context - perhaps if preferences is not the recommended approach we should use `Global under extensions`  -  but this could be a separate PR, no ?

pdettori:
As for the mapping between control plane names and context names - if we move to global the `kflex-initial-ctx-name` it should look like something like this:

```yaml
- context:
    cluster: kflex-cp1-cluster
    extensions:
    - extension:
        data:
          kflex-controlplane-name: kflex-cp1    # associated controlplane
        metadata:
          creationTimestamp: null
      name: kubeflex
    user: kflex-cp1-admin
  name: kflex-cp1
 ```

pdettori:
need also to think what happens if that extension data is lost or corrupted (in the past we saw this problem with losing `kflex-initial-ctx-name` - we has also to add a command that allowed to user to manually set the initial context when that happened :
`kflex ctx --set-current-for-hosting` - need probably a way here too to allow to recover lost mappings without having to manually edit the yaml

rxinui:
Sure, we can proceed that way if you agree with my approach. Sorry I'm aware preferences makes things work but if we can make kflex follows best practices (and I hope I'm right) then it seems a good direction to follow.

I'll make a follow up PR for the change then.

Regarding the initial context, I think I was unclear. Each context stores its initial context which is the control plane name as of now. I can remove it if necessary but it has nothing to do with "initial context" you were referring to.

In a way, having the control plane name associated to a context is enough to determine the initial context name because the policy is "certs.GenerateContextName(cpName)" which returns cpName. However, the intent is implicit and may change in the future. Say, we decide to append "-ctx-<uuid>" as a result of "certs.GenerateContextName(cpName)" then we can't trace back the first context name. Hence, my proposition of storing both "controlplane-name" and "kflex-initial-ctx-name" under each context's extension

pdettori (commented):
PR overall looks good. I am fine with keeping the current proposed context-based extension, I just suggest using  `controlplane-name` and `initial-ctx-name` instead of the names with `kubeflex-` - this is redundant as the extension is already named   `kubeflex` . Also, I prefer to avoid using the same name used by the other extension for keeping the hosting cluster context name, which is currently also named `kflex-initial-ctx-name` but it has a completely different meaning - it represents the context associated with the cluster hosting kubeflex.

rxinui:
Totally agree. I will change the naming convention and ping you back.

rxinui:
@pdettori in the meantime, could you take a look at the following discussions, I would like your opinion and whether or not I should fix it in this or next PR.

1. https://github.com/kubestellar/kubeflex/blob/a1fdc2bec487364ac60bd22c96bcece3f435fe7b/pkg/kubeconfig/kubeconfig.go#L157-L176
2. https://github.com/kubestellar/kubeflex/blob/a1fdc2bec487364ac60bd22c96bcece3f435fe7b/pkg/kubeconfig/kubeconfig.go#L288C1-L302C2

pdettori:
> should we restrict the usage of `kflex ctx`  to ONLY controlplane managed by kflex ?? 

Yes! That would be my take now

PR #382, opened 2025-05-20
✨ removing lint and verify - they are not available in makefile

clubanderson:
## Related issue(s)

Fixes #


clubanderson:
sorry, was having a heck of a time with my signoff and signing tonight.

clubanderson:
if you can add lint to kubeflex, that would be a great next step in CI rigor

pdettori:
@clubanderson - thanks, will look into it

PR #371, opened 2025-05-18
✨ feature: Allow multiple PostCreateHooks when creating a control plane

asmit27rai:
## Summary
- Replaces single hook fields with slice of `PostCreateHookUse` structs
- Adds CLI support for multiple `-p/--post-create-hook` flags
- Implements namespaced variables using `--set hookName.key=value` syntax
- Updates reconciler logic to handle multiple hooks with proper status tracking
- Maintains backward compatibility with status tracking system

## Related issue(s)
Fixes: #225 

## Testing
- Updated e2e tests to verify multi-hook scenarios
- Added validation for hook variable parsing
- Verified CRD upgrade path from previous versions

PR #384, opened 2025-05-20
feature: implemented multiple PostCreateHooks with backward compatibility

### LINKED ISSUE #225: feature: Allow more than one PostCreateHook when creating a control plane
### Feature Description

Currently a control plane construction can invoke at most one PostCreateHook. This feature would allow more than one.

This would be very helpful in untangling dependencies among various repos in the KubeStellar organization. Currently, wherever more than one thing has to be added to a control plane, something has to define one PostCreateHook that adds all of it. Consider, for example, a WDS: it needs both the kubestellar controller-manager from the kubestellar repo and will soon also need the transport controller from the ocm-transport-plugin repo.

### Proposed Solution

In `ControlPlaneSpec`, replace

```go
	PostCreateHook     *string           `json:"postCreateHook,omitempty"`
	PostCreateHookVars map[string]string `json:"postCreateHookVars,omitempty"`
```

with

```go
	PostCreateHooks     []PostCreateHookUse           `json:"postCreateHooks,omitempty"`
```

and define

```go
type PostCreateHookUse struct {
    HookName *string `json:"hookName"`
    Vars map[string]string `json:"vars,omitempty"`
}
```

In the CLI, allow multiple `-p --set ... --set` trains in series.

### Want to contribute?

### Additional Context

_No response_

asmit27rai:
# Summary

## In `api/v1alpha1/controlplane_types.go`
 - Add PostCreateHooks field for multiple hooks
 - Deprecate legacy single hook fields
 - Add PostCreateHookUse struct

## In `pkg/reconcilers/shared/postcreate_hook.go`
 - Process both legacy and new hooks
 - Handle multiple hooks in sequence
 - Update status per hook

## Backward Compatibility Strategy
**Old Fields**: Continue to work as before but are marked deprecated.

**New Fields**: Allow multiple hooks via `PostCreateHooks`.

**Reconciler**: Processes both old and new fields, merging them into a single list.

This ensures existing users using `postCreateHook` and `postCreateHookVars` are unaffected, while new users can adopt `postCreateHooks`. The CLI can be updated separately to support multiple hooks via new flags while maintaining backward compatibility with the old ones.

Fixes #225

### DIFF
--- a/pkg/reconcilers/shared/postcreate_hook.go
+++ b/pkg/reconcilers/shared/postcreate_hook.go
@@ -45,72 +45,82 @@ func (r *BaseReconciler) ReconcileUpdatePostCreateHook(ctx context.Context, hcp 
     logger := clog.FromContext(ctx)
     namespace := util.GenerateNamespaceFromControlPlaneName(hcp.Name)
     
-    if hcp.Spec.PostCreateHook == nil {
-        return nil
-    }
-
-    // Hook is only applied once after CP creation
-    _, ok := hcp.Status.PostCreateHooks[*hcp.Spec.PostCreateHook]
-    if ok {
-        return nil
-    }
-
-    logger.Info("Running ReconcileUpdatePostCreateHook", "post-create-hook", *hcp.Spec.PostCreateHook)
-
-    // Get the post create hook
-    hook := &v1alpha1.PostCreateHook{
-        ObjectMeta: metav1.ObjectMeta{
-            Name: *hcp.Spec.PostCreateHook,
-        },
-    }
-    if err := r.Client.Get(context.TODO(), client.ObjectKeyFromObject(hook), hook, &client.GetOptions{}); err != nil {
-        return fmt.Errorf("error retrieving post create hook %s: %w", *hcp.Spec.PostCreateHook, err)
-    }
-
-    // Build variables with proper precedence:
-    // 1. Default vars from PostCreateHook
-    // 2. User-provided vars from ControlPlane (override defaults)
-    // 3. Built-in system vars (highest priority)
-    vars := make(map[string]interface{})
+    // Collect all hooks to process (legacy + new)
+    var hooks []v1alpha1.PostCreateHookUse
     
-    // Add default variables from hook spec
-    for _, dv := range hook.Spec.DefaultVars {
-        vars[dv.Name] = dv.Value
+    // Add legacy hook if specified
+    if hcp.Spec.PostCreateHook != nil {
+        hooks = append(hooks, v1alpha1.PostCreateHookUse{
+            HookName: hcp.Spec.PostCreateHook,
+            Vars:     hcp.Spec.PostCreateHookVars,
+        })
     }
     
-    // Override with user-provided variables from control plane
-    for key, val := range hcp.Spec.PostCreateHookVars {
-        vars[key] = val
-    }
+    // Add new hooks
+    hooks = append(hooks, hcp.Spec.PostCreateHooks...)
     
-    // Add system variables (highest priority)
-    vars["Namespace"] = namespace
-    vars["ControlPlaneName"] = hcp.Name
-    vars["HookName"] = *hcp.Spec.PostCreateHook
-
-    // Apply the hook templates
-    if err := applyPostCreateHook(ctx, r.ClientSet, r.DynamicClient, hook, vars, hcp); err != nil {
-        if util.IsTransientError(err) {
-            return fmt.Errorf("failed to apply post-create hook: %w", err) // Retry
+    for _, hook := range hooks {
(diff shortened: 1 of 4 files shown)


asmit27rai:
@pdettori I closed #371  due to many unnecessary commits..
This is revised form of it...

pdettori:
@asmit27rai - thanks, I will review - in the meantime, could you pls. set the correct PR type? The CI check is failing. It should be ✨ feature:

pkg/reconcilers/shared/postcreate_hook.go
```diff
@@ -45,72 +45,82 @@ func (r *BaseReconciler) ReconcileUpdatePostCreateHook(ctx context.Context, hcp
...
     
-    // Add default variables from hook spec
-    for _, dv := range hook.Spec.DefaultVars {
-        vars[dv.Name] = dv.Value
+    // Add legacy hook if specified
+    if hcp.Spec.PostCreateHook != nil {
+        hooks = append(hooks, v1alpha1.PostCreateHookUse{
+            HookName: hcp.Spec.PostCreateHook,
+            Vars:     hcp.Spec.PostCreateHookVars,
+        })
```
pdettori, line 56: If the same hook is specified both as the legacy field and in the new array, it may be applied twice unless handled elsewhere.  Add a check to avoid duplicate execution of hooks (e.g., by hook name).

pkg/reconcilers/shared/postcreate_hook.go
```diff
@@ -45,72 +45,82 @@ func (r *BaseReconciler) ReconcileUpdatePostCreateHook(ctx context.Context, hcp
...
+        // Skip already applied hooks
+        if hcp.Status.PostCreateHooks != nil && hcp.Status.PostCreateHooks[hookName] {
+            continue
+        }
+        
+        logger.Info("Processing post-create hook", "hook", hookName)
+        
+        // Get hook definition
+        pch := &v1alpha1.PostCreateHook{}
+        if err := r.Client.Get(ctx, client.ObjectKey{Name: hookName}, pch); err != nil {
```
pdettori, line 78: On transient errors, the function immediately returns, preventing other hooks from being processed.  Would be better accumulating errors and reporting them collectively, or ensure that a failed hook (transient or not) doesn’t block other hooks from being attempted

test/e2e/setup-kubeflex.sh
```diff
@@ -24,6 +24,29 @@ SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &> /dev/null && pwd)
...
+fi
+
+# Install ingress-nginx using Helm
+helm repo add ingress-nginx https://kubernetes.github.io/ingress-nginx
+helm install ingress-nginx ingress-nginx/ingress-nginx \
+  --namespace ingress-nginx \
+  --create-namespace \
+  --version 4.10.0 \
+  --set controller.service.type=NodePort \
+  --set controller.hostPort.enabled=true
```
pdettori, line 42: the script installs ingress-nginx twice: once via Helm and once via kubectl apply. This may lead to race conditions or resource conflicts. Use only one method to install ingress-nginx.

francostellari:
Is there a documentation update that goes along with the Go code change?
I would like to understand how to use the new hook mechanism from KS core-chart.
@Rupam-It can you please monitor this and propose an update of the core-chart when the time comes?

francostellari:
@asmit27rai when multiple hooks are listed... is there a consistent order by which they are executed? perhaps the order by which they are listed?

asmit27rai:
> To be fully clear... this older YAML:
> 
> ```yaml
>   postCreateHook: wds
>   postCreateHookVars:
>     ITSName: "{{ $cp.ITSName | default "" }}"
>     APIGroups: "{{ $cp.APIGroups | default "" }}"
> ```
> 
> could be changed to this YAML after this PR merges:
> 
> ```yaml
>   postCreateHooks:
>   -  hookName: wds
>      vars:
>        ITSName: "{{ $cp.ITSName | default "" }}"
>       APIGroups: "{{ $cp.APIGroups | default "" }}"
> ```
> 
> `vars` is a dictionary like before or does it need to be a list?

The new `vars` field is still a dictionary (map), not a list.
You can confirm it here.

```bash
type Var struct {
	Name  string `json:"name"`
	Value string `json:"value"`
}
```

asmit27rai:
> Since the new `vars` seems to be local and limited to the specific hook, can the existing `postCreateHookVars` be used to set "global" variables for all the hooks?

- The legacy `postCreateHookVars` is only applied to the legacy single hook (when using `postCreateHook`)

- It is not shared with hooks defined in `postCreateHooks`

- Each hook in `postCreateHooks` has its own isolated vars dictionary

PR #386, opened 2025-06-06
✨ Usage of Extensions

### LINKED ISSUE #385: feature: move away from Preferences to Extensions
### Feature Description

**BREAKING CHANGES**

## Context

Currently, hosting cluster context is set within `preferences[*].extensions` of a kubeconfig.

However, Kubeconfig file recommends the usage of `extensions` to store metadata for 3rd-party as `preferences` refers to an end-user preference.

## New config management

We are leveraging 2 type of extensions, under the key `kubeflex`:

1. `extensions.kubeflex:` which is global to the kubeconfig
2. `contexts[].extensions.kubeflex:` which is local to a context

Information related to a context,controlplane name associated to a context, are stored in 2. while information that are general to kubeflex regardless of a context are stored in 1.

### Want to contribute?

### Additional Context

These changes have been decided with @pdettori

rxinui:
## Summary

The implementation of `Preferences` used to be *on-the-fly* by using `corev1.ConfigMap` as "k8s.io/client-go/tools/clientcmd/api" requires `runtime.Object` type for `Extension`.

In our new implementation, a new file `extensions.go` is created where structs are defined. This allow us Static typing and check at compilation time, reducing likelihood of errors.

## BREAKING CHANGES

Within the code, breaking changes are indicated through comment `// BREAKING CHANGE`

```go
const (
	ExtensionConfigName                = "config-extension-name"    // BREAKING CHANGE
	ExtensionHostingClusterContextName = "hosting-cluster-ctx-name" // BREAKING CHANGE
```

Those values have changed. Also, perhaps some of these values are now useless. In fact, all kubeflex metadata are stored under `kubeflex` key (ie. `extensions.kubeflex`) therefore `ExtensionConfigName` seems irrelevant.

Functions defined in `pkg/kubeconfig/kubeconfig.go` are no longer looking up to `preferences.extensions` within a Kubeconfig. Thereby, any clients/end-users using kflex version prior this PR has their Kubeconfig populated by `Preferences`. Once this PR merge to `kubestellar/kubeflex`, metadata stored under `Preferences` are ignored (hence become irrelevant) and new metadata will be inserted under `extensions` and `contexts[].extensions`. Breaking changes are to be expected for the end user.

Old way:
```yaml
preferences:
  extensions:
  - extension:
      data:
        kflex-initial-ctx-name: kind-kubeflex    # key=ExtensionHostingClusterContextName
      metadata:
        creationTimestamp: null
        name: kflex-config-extension-name        # value=ExtensionConfigName
    name: kflex-config-extension-name            # value=ExtensionConfigName
```

New way:
```yaml
# global
extensions:
- extension:
    data:
      hosting-cluster-ctx-name: kind-kubeflex    # key=ExtensionHostingClusterContextName
    metadata:
      creationTimestamp: null
      name: kubeflex                             # value=ExtensionKubeflexKey
  name: kubeflex                                 # value=ExtensionKubeflexKey
# ...
# context-scope
contexts:
- name: kind-kubeflex
  context:
    cluster: kind-kubeflex
    user: kind-kubeflex
  extensions:
  - extension:
      data:
        first-ctx-name: kind-kubeflex            # key=ExtensionInitialContextName
        is-hosting-cluster-ctx: true             # key=ExtensionContextsIsHostingCluster
      metadata:
        creationTimestamp: null
        name: kubeflex
    name: kubeflex
```

To alleviate those breaking changes, a follow up PR is to be considerated. The creation of a command `kflex config` seems ideal where subcommands interact with Kubeconfig. For instance, to help the transition to this breaking changes, we can imagine:
- `kflex config tidy` to migrate/clean-up a kubeconfig to adapt with kubeflex latest changes.

## Related issue(s)

Fixes #385

### DIFF
--- a/cmd/kflex/main.go
+++ b/cmd/kflex/main.go
@@ -19,6 +19,7 @@ package main
 import (
 	"os"
 
+	"github.com/fatih/color"
 	"github.com/kubestellar/kubeflex/cmd/kflex/adopt"
 	"github.com/kubestellar/kubeflex/cmd/kflex/common"
 	"github.com/kubestellar/kubeflex/cmd/kflex/create"
@@ -30,12 +31,6 @@ import (
 	"github.com/spf13/cobra"
 	"k8s.io/client-go/tools/clientcmd"
 )
-
-// Version injected by makefile:LDFLAGS
-var Version string
-
-// BuildDate injected by makefile:LDFLAGS
-var BuildDate string
 
 var rootCmd = &cobra.Command{
 	Use:   "kflex",
@@ -58,8 +53,12 @@ func init() {
 }
 
 // TODO - work on passing the verbosity to the logger
-
 func main() {
+	// TODO - find a way to inject it using Makefile
+	common.WarningMessage = "WARNING: current kflex version introduces BREAKING CHANGES related to kflex and your kubeconfig file which may interrupt kflex to function properly.\nSee https://github.com/kubestellar/kubeflex/blob/main/docs/users.md"
+	if common.WarningMessage != "" {
+		color.Yellow(common.WarningMessage)
+	}
 	if err := rootCmd.Execute(); err != nil {
 		os.Exit(1)
 	}
--- a/docs/users.md
+++ b/docs/users.md
@@ -1,5 +1,54 @@
 # User's Guide
 
+## Breaking changes
+
+### v0.9.0 
+
+Kubeflex configuration is stored within Kubeconfig file. Prior this version, `kflex` put its configuration under 
+
+```yaml
+preferences:
+  extensions:
+  - extension:
+      data:
+        kflex-initial-ctx-name: kind-kubeflex  # indicates to kflex the hosting cluster context
+      metadata:
+        creationTimestamp: null
+        name: kflex-config-extension-name
+    name: kflex-config-extension-name          # indicates to kflex that this extension belongs to it
+```
+
+Starting `v0.9.0`, configuration remains within kubeconfig file but leverages `extensions:` and context-scope extension. For instance, the previous example would be translated as follow:
+
+```yaml
+extensions:                                          # change -> no longer preferences. See https://kubernetes.io/docs/reference/config-api/kubeconfig.v1/#Config
+  - extension:
+      data:
+        hosting-cluster-ctx-name: kind-kubeflex      # change -> key name "hosting-cluster-ctx-name"
+      metadata:
+        creationTimestamp: 2025-06-09T22:36:17+02:00 # creation timestamp using ISO 8601 seconds
+        name: kubeflex                               # same as below
+    name: kubeflex                                   # change -> new extension name "kubeflex"
+# ...
+# Find the context corresponding to "hosting-cluster-ctx-name" (here "kind-kubeflex")
+contexts:
+  - context:
+      cluster: kind-kubeflex
+      user: kind-kubeflex
+      # add extensions below and its information
+      extensions:
+      - extension:
+        data:
+          is-hosting-cluster-ctx: true                 # change -> key name "is-hosting-cluster-ctx" with "true"
+        metadata:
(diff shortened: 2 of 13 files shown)


rxinui:
CC: @pdettori @dumb00002

rxinui:
The constant `ExtensionConfigName` does not bring any value in the new way and in my opinion, should be totally removed as `ExtensionKubeflexKey` is taking over indicating clearly that such extension refers to kubeflex usage.

pkg/kubeconfig/extensions.go
```diff
@@ -0,0 +1,183 @@
...
+		return err
+	}
+	fmt.Printf("dataJson: %s\n", dataJson)
+	err = json.Unmarshal(dataJson, &receiver)
+	fmt.Printf("unmarshal: %v\n", *receiver)
+	if err != nil {
+		return err
+	}
+	return nil
+}
```
pdettori, line 183: I think the code assumes all kubeconfig for Kubeflex have well-formed KubeflexExtensions objects under the extensions key. If a user manually edits kubeconfig, or if there are partial/invalid extension values, code may panic or behave unexpectedly.

pkg/kubeconfig/kubeconfig.go
```diff
@@ -286,81 +255,63 @@ func RenameKey(m interface{}, oldKey string, newKey string) {
...
+	kconf.Contexts[hostingContext].Extensions, err = kflexContextConfig.ParseToKubeconfigExtensions()
+	if err != nil {
+		return fmt.Errorf("error while setting hosting cluster context to context extensions: %v", err)
+	}
+	return nil
 }
 
+// TODO: the signature is confusing. It switches context but expect controlPlane name as parameter
+// NOTE: Perhaps SwitchContext should only switch context to a context name
+// NOTE: Create a new function SwitchToControlPlaneContext to find a context using control plane name
```
pdettori, line 287: Agreed - that would be more clear

pdettori (commented):
Overall looks good, I have couple of general comments I think would be best to address before merging:

1. Since this is a breaking change, it would be helpful adding a script or command to migrate old kubeconfig extensions to the new format, or at least provide documentation about this breaking change and how the user can migrate (even by manually editing the file if no automation is provided)

2. Should add some text to README or  release notes to explain the new extension mechanism, breaking changes, and manual steps required by users.

pkg/kubeconfig/extensions.go
rxinui: True. What should we do then? Isn't a `panic` a good situation to act as such?

docs/users.md
```diff
@@ -1,5 +1,54 @@
 # User's Guide
 
+## Breaking changes
```
rxinui, line 3: @pdettori I have added this section to describe the breaking changes what needs to be done by the end-user.

cmd/kflex/main.go
```diff
@@ -58,8 +53,12 @@ func init() {
 }
 
 // TODO - work on passing the verbosity to the logger
-
 func main() {
+	// TODO - find a way to inject it using Makefile
+	common.WarningMessage = "WARNING: current kflex version introduces BREAKING CHANGES related to kflex and your kubeconfig file which may interrupt kflex to function properly.\nSee https://github.com/kubestellar/kubeflex/blob/main/docs/users.md"
```
rxinui, line 58: Everytime `kflex` is ran, a YELLOW WARNING message will be displayed to indicate the user about breaking changes, with a URL pointing to the end-user documentation.

Ideally, I wanted to inject the message within the `Makefile` but it is rather complicated to do it. So I fallback in Go for now.

Perhaps it is a sign to move to https://taskfile.dev/usage/

pdettori: great, thx

pkg/kubeconfig/extensions.go
pdettori: yes, as long as the user understands what caused it

PR #390, opened 2025-06-10
✨ Add `kflex config set-hosting`

### LINKED ISSUE #387: feature: enable set hosting cluster ctx from CLI
### Feature Description

> need also to think what happens if that extension data is lost or corrupted (in the past we saw this problem with losing kflex-initial-ctx-name - we has also to add a command that allowed to user to manually set the initial context when that happened :
> kflex ctx --set-current-for-hosting - need probably a way here too to allow to recover lost mappings without having to manually edit the yaml

by @pdettori 

### Proposed Solution

There are multiple solution

@paolo suggested to extend the `ctx` command with an option `--set-current-for-hosting` to define the current context as hosting context in the kubeconfig.

@rxinui suggests 2 approaches
1. Create a subcommand `kflex ctx set-hosting [CTX_NAME]`. This approach dedicates a subcommand to that scenario and allow to specify a context by its name. If `CTX_NAME` not provided, then use the current context.

2. Create a subcommand `kflex ctx set CONFIG_KEY=CONFIG_VALUE [...CONFIG_KEY=CONFIG_VALUE]`. This approach is broader and extends the capability of `kflex ctx` to future needs (ie. set different controlplane name for a context because of data corruption).

### Want to contribute?

### Additional Context

_No response_

rxinui:
## Summary

Introducing `kflex config` command to interact with kubeconfig/kubeflex configuration.

This PR introduces the first config feature `set-hosting` to set/change the value of hosting cluster context name.

**Usage:** `kflex config set-hosting $mycp` 

Set hosting cluster context name to `$mycp` value

## Related issue(s)

Fixes #387

### DIFF
--- a/cmd/kflex/config/config.go
+++ b/cmd/kflex/config/config.go
@@ -22,7 +22,7 @@ import (
 
 func Command() *cobra.Command {
 	command := &cobra.Command{
-		Use:   "config",
+		Use:   "config COMMAND",
 		Short: "Configure kubeflex",
 		Long:  `Configure settings of kubeflex set within a kubeconfig file`,
 		Args:  cobra.ExactArgs(1),
--- a/cmd/kflex/config/config.go
+++ b/cmd/kflex/config/config.go
@@ -1,0 +1,32 @@
+/*
+Copyright 2023 The KubeStellar Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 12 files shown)


rxinui:
@pdettori i do not know what you aren't automatically set as a reviewer, but here the last PR before internship

rxinui:
@pdettori thank you for your deep review, very much appreciated. I have updated it accordingly.

pdettori:
/lgtm
thanks @rxinui !

Issue #394, opened 2025-06-12
feature:


francostellari:
### Feature Description

Would it be possible for kflex to 
1. determine when a PCH is completed/ready state, i.e. all listed resources are created and pods/deployments/jobs are in corresponding ready/completed state
2. add a new status filed to the corresponding CP to show `PostCreateHookCompleted = T/F`?

### Proposed Solution

The controller would have to analyze the PCH template yaml and monitor the life of the various objects

### Want to contribute?

- [ ] I would like to work on this issue.

pdettori:
my suggestion is to start simple and just support status checks for the most common cases - jobs, deployments, statefulsets

francostellari:
@rupam-it are you app to give this a shot?

@pdettori my initial reference to #383 was related to the fact that I'm unclear if a similar mechanism is already being implemented by that PR... so may be can use it.

PR #397, opened 2025-06-16
✨ add status of postcreatehooks completion!

Rupam-It:
## Related issue(s)

Fixes #
as of now I"m working on this issue #394  !
and done some work but I have to test it!

### DIFF
--- a/api/v1alpha1/controlplane_types.go
+++ b/api/v1alpha1/controlplane_types.go
@@ -32,12 +32,27 @@ type ControlPlaneSpec struct {
 	// +optional
 	// +kubebuilder:default:=31536000
 	TokenExpirationSeconds *int64            `json:"tokenExpirationSeconds,omitempty"`
+	// Keep the deprecated fields for backwards compatibility
+	// Deprecated: Use PostCreateHooks instead
 	PostCreateHook         *string           `json:"postCreateHook,omitempty"`
+	// Deprecated: Use PostCreateHooks instead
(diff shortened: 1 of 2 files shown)


francostellari:
@Rupam-It can you show a working demo of the new feature using a dummy PCH with a known sleep time? Thank you

francostellari:
@pdettori can you take a look and give @Rupam-It some suggestions? Thank you

pdettori:
> Do we need to add a bool WaitForPostCreateHooks (default false) to the CP CRD to enable/disable the new behavior for backwards compatibility?

@francostellari - yes, that would be my preference as well

Rupam-It:
> @Rupam-It - I think it would help to add a e2e test (similar to what you show here in the demo) - see other tests in `test/e2e`

I'm working on that and most of them are resolve already !
one things to confirm is that 
```
1. Create ControlPlane
2. API Server starts → becomes ready
3. PostCreateHook gets applied → postCreateHooks.hook-name: true
4. PostCreateHook resources (Job, Deployment) → become ready
5. postCreateHookCompleted: true
6. FINALLY → ControlPlane Ready: True
```
this is the functionality we want to achieve right?
cc @francostellari

francostellari:
@Rupam-It I think your workflow looks ok but we want to add a bool `WaitForPostCreateHooks` (default false) to the CP CRD to enable/disable the new behavior for backwards compatibility

PR #445, opened 2025-06-29
📖 answer of why postgresql is a post create hook job

### LINKED ISSUE #401: feature: document why postgresql is not a subchart
### Feature Description

As someone debugging problems with the kubeflex Helm chart (e.g., #400), I wonder why the postgresql chart is instantiated with a post-install hook Job rather than by being a sub-chart.

Part (maybe the whole, I do not know!) of the answer is that the kubeflex chart is designed to make the instantiation of the postgresql chart optional. But that begs the question, why is instantiation of the postgresql chart optional?

Also, why does the postgresql chart get instantiated into a fixed namespace rather than the kubeflex chart's namespace?

### Proposed Solution

Document the answers. Or simplify the system.

### Want to contribute?

### Additional Context

_No response_

Rupam-It:
## Related issue(s)

Fixes #401


pdettori (commented):
/lgtm
/approve

PR #452, opened 2025-07-04
✨ Add guard for kflex ctx delete: confirmation for non-KubeFlex contexts

### LINKED ISSUE #392: feature: guard irreversible `kflex ctx delete`
### Feature Description

`kflex ctx` is a extended and shortcut to `kubectl` operation on contexts.

Kubeflex enable the deletion of a context by using `kflex ctx delete $ctxname`. As of now, it is also possible to delete context that are not managed by kubeflex (ie. ocp, ocm...)

Since the deletion is done using `kflex`, it would be pertinent and safer to:
- Delete context that are managed by kubeflex using `kflex ctx delete`
- Delete context that are NOT managed by kubeflex AFTER asking confirmation and display a warning message indicating that the given context is not managed by kflex

This prevent mistake deletions, which are irreversible.

### Proposed Solution

Create a guard when `kflex ctx delete` is invoked

If a context is not managed by `kflex`, a guard should be defined. It asks the end-user a confirmation of its action.

The guard can expected to distinct user-prompt:
1. Inform the user that the following context is not managed by `kflex`
2. Ask confirmation for deletion (this implies interactive prompt and/or forcing flag)

### Want to contribute?

### Additional Context

Following breaking changes https://github.com/kubestellar/kubeflex/blob/main/docs/users.md#v090 - the existence of an context-scoped extension named `kubeflex` (ie. `name: kubeflex`) indicates that the given context is managed by `kflex`.

rishi-jat:
Summary :-

This PR implements a safety guard for the kflex ctx delete command to prevent accidental deletion of Kubernetes contexts not managed by KubeFlex.

Changes
	•	Added IsContextManagedByKubeflex function to detect if a context is managed by KubeFlex.
	•	Modified ExecuteCtxDelete to:
	•	Display a warning if the context is not managed by KubeFlex.
	•	Prompt the user for confirmation before deleting non-managed contexts.
	•	Support a --force flag to bypass confirmation for automation.
	•	Updated tests to cover both KubeFlex-managed and non-managed contexts.

closes #392

### DIFF
--- a/cmd/kflex/ctx/delete.go
+++ b/cmd/kflex/ctx/delete.go
@@ -31,6 +31,19 @@ import (
 	"github.com/spf13/cobra"
 )
 
+type DeleteOptions struct {
+	Force bool
+}
+
+type DeleteOption func(*DeleteOptions)
+
+// WithForce sets the force option to bypass confirmation
+func WithForce() DeleteOption {
+	return func(opts *DeleteOptions) {
+		opts.Force = true
+	}
+}
+
 func CommandDelete() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "delete CONTEXT",
@@ -43,7 +56,13 @@ func CommandDelete() *cobra.Command {
 			chattyStatus, _ := cmd.Flags().GetBool(common.ChattyStatusFlag)
 			force, _ := cmd.Flags().GetBool("force")
 			cp := common.NewCP(kubeconfig)
-			return ExecuteCtxDelete(cp, args[0], chattyStatus, force)
+
+			var options []DeleteOption
+			if force {
+				options = append(options, WithForce())
+			}
+
+			return ExecuteCtxDelete(cp, args[0], chattyStatus, options...)
 		},
 	}
 	cmd.Flags().BoolP("force", "f", false, "Force deletion without confirmation")
@@ -51,7 +70,12 @@ func CommandDelete() *cobra.Command {
 }
 
 // Execute kflex ctx delete
-func ExecuteCtxDelete(cp common.CP, ctxName string, chattyStatus bool, force bool) error {
+func ExecuteCtxDelete(cp common.CP, ctxName string, chattyStatus bool, options ...DeleteOption) error {
+	opts := &DeleteOptions{}
+	for _, option := range options {
+		option(opts)
+	}
+
 	var wg sync.WaitGroup
 	done := make(chan bool)
 
@@ -60,23 +84,21 @@ func ExecuteCtxDelete(cp common.CP, ctxName string, chattyStatus bool, force boo
 		return fmt.Errorf("error loading kubeconfig: %v", err)
 	}
 
-	// Check if context is managed by KubeFlex
-	if !kubeconfig.IsContextManagedByKubeflex(kconf, ctxName) {
-		if !force {
-			fmt.Printf("Warning: Context '%s' is not managed by KubeFlex.\n", ctxName)
-			fmt.Print("Are you sure you want to delete this context? (y/N): ")
+	// Check if context is managed by KubeFlex and force is not set
+	if !kubeconfig.IsContextManagedByKubeflex(kconf, ctxName) && !opts.Force {
+		fmt.Printf("Warning: Context '%s' is not managed by KubeFlex.\n", ctxName)
+		fmt.Print("Are you sure you want to delete this context? (y/N): ")
 
-			reader := bufio.NewReader(os.Stdin)
-			response, err := reader.ReadString('\n')
-			if err != nil {
-				return fmt.Errorf("error reading user input: %v", err)
-			}
+		reader := bufio.NewReader(os.Stdin)
+		response, err := reader.ReadString('\n')
+		if err != nil {
+			return fmt.Errorf("error reading user input: %v", err)
+		}
 
-			response = strings.TrimSpace(strings.ToLower(response))
-			if response != "y" && response != "yes" {
-				fmt.Println("Deletion cancelled.")
-				return nil
-			}
+		response = strings.TrimSpace(strings.ToLower(response))
+		if response != "y" && response != "yes" {
+			fmt.Println("Deletion cancelled.")
+			return nil
 		}
 	}
 
--- a/cmd/kflex/ctx/delete.go
+++ b/cmd/kflex/ctx/delete.go
@@ -17,7 +17,10 @@ limitations under the License.
 package ctx
 
 import (
+	"bufio"
 	"fmt"
+	"os"
+	"strings"
 	"sync"
 
 	// "sync"
@@ -29,7 +32,7 @@ import (
 )
 
 func CommandDelete() *cobra.Command {
-	return &cobra.Command{
+	cmd := &cobra.Command{
 		Use:   "delete CONTEXT",
 		Short: "Delete a context",
 		Long:  `Delete a context in the kubeconfig file`,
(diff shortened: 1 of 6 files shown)


cmd/kflex/ctx/delete.go
```diff
@@ -38,21 +41,47 @@ func CommandDelete() *cobra.Command {
...
+			return ExecuteCtxDelete(cp, args[0], chattyStatus, force)
 		},
 	}
+	cmd.Flags().BoolP("force", "f", false, "Force deletion without confirmation")
+	return cmd
 }
 
 // Execute kflex ctx delete
-func ExecuteCtxDelete(cp common.CP, ctxName string, chattyStatus bool) error {
+func ExecuteCtxDelete(cp common.CP, ctxName string, chattyStatus bool, force bool) error {
```
rxinui, line 54: since `--force` is an optional flag, i would like you to implement the golang option pattern to handle this.

See how i implemented it https://github.com/rishi-jat/kubeflex/blob/21c4b31bd1efad3923707f29c72a991d959d5002/cmd/kflex/common/cp.go#L33C1-L55C2

Also, see https://golang.cafe/blog/golang-functional-options-pattern.html

cmd/kflex/ctx/delete_test.go
```diff
@@ -71,11 +71,43 @@ func TestDeleteNonExistentContext(t *testing.T) {
 		t.Errorf("error loading kubeconfig: %v", err)
 	}
 	nCtx := len(kconf.Contexts)
-	err = ExecuteCtxDelete(cp, noneCtxName, false)
+	err = ExecuteCtxDelete(cp, noneCtxName, false, false)
```
rxinui, line 74: same comment, see above

cmd/kflex/ctx/delete_test.go
```diff
@@ -71,11 +71,43 @@ func TestDeleteNonExistentContext(t *testing.T) {
...
+
+// Test delete a context that is not managed by KubeFlex
+func TestDeleteNonKubeflexContext(t *testing.T) {
+	ctxName := "cptobedeleted"
+	setupMockContextWithoutKubeflex(t, kubeconfigPath, ctxName)
+	defer teardown(t, kubeconfigPath)
+
+	// Start test
+	cp := common.NewCP(kubeconfigPath, common.WithName(ctxName))
+	err := ExecuteCtxDelete(cp, ctxName, false, true)
```
rxinui, line 91: same comment, see above

rxinui (changes requested):
Some changes to do, otherwise good job, thank you for keeping the codebase flow as it is.

rishi-jat:
Hi @rxinui, the requested changes are done. Please have a look when you get a chance.

cmd/kflex/ctx/delete.go
```diff
@@ -38,21 +54,56 @@ func CommandDelete() *cobra.Command {
...
+			force, _ := cmd.Flags().GetBool("force")
 			cp := common.NewCP(kubeconfig)
-			return ExecuteCtxDelete(cp, args[0], chattyStatus)
+
+			var options []DeleteOption
+			if force {
+				options = append(options, WithForce())
+			}
+
+			return ExecuteCtxDelete(cp, args[0], chattyStatus, options...)
```
rxinui, line 65: Very good, thank you

rishi-jat:
Hi @pdettori ,
Reason for test failure:
The E2E test was failing because, after renaming a kube context, it lost the KubeFlex metadata. As a result, the delete command prompted for confirmation (since it didn’t recognize the context as KubeFlex-managed), which caused the test to hang in CI.
What I did to fix it:
Updated the context rename logic to correctly preserve KubeFlex extensions/metadata.
Modified the E2E test to use the --force flag for context deletion, preventing interactive prompts in CI.
With these changes, renamed contexts are now properly recognized as KubeFlex-managed, and the E2E tests pass in CI.
Let me know if you need any more details!

cmd/kflex/ctx/rename.go
```diff
@@ -62,10 +63,47 @@ func ExecuteCtxRename(cp common.CP, ctxName string, newCtxName string, toSwitch
...
+					newContext.Extensions[key] = value
+				}
+			} else {
+				// For non-KubeFlex extensions, copy as-is
+				newContext.Extensions[key] = value
+			}
+		}
+	}
+
+	kconf.Contexts[newCtxName] = newContext
```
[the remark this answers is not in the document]
rxinui: if the code has to be changed, the only thing to do is to move from

```
kconf.Contexts[newCtxName] = &api.Context{
		Cluster:  newClusterName,
		AuthInfo: newAuthInfoAdminName,
	}
```

to 
```
kconf.Contexts[newCtxName] = &api.Context{
		Cluster:  newClusterName,
		AuthInfo: newAuthInfoAdminName,
                Extensions: kconf.Contexts[ctxName].Extensions // make sure it is a copy of its value
	}
```

rishi-jat:
@rxinui All requested changes are done.
The context rename now only copies the Extensions as-is, with no modification to extension data.
Let me know if anything else is needed—thanks for your guidance!

rishi-jat:
Hi @rxinui,
I’ve added a unit test to ensure that kflex ctx delete prompts a warning and confirmation when deleting a context not managed by KubeFlex, exactly as you requested.

cmd/kflex/ctx/delete_test.go
```diff
@@ -17,65 +17,136 @@ limitations under the License.
...
+	go func() {
+		w.Write([]byte("n\n"))
+		w.Close()
+	}()
+
+	// Capture output
+	oldStdout := os.Stdout
+	rOut, wOut, _ := os.Pipe()
+	os.Stdout = wOut
+	defer func() { os.Stdout = oldStdout }()
```
rxinui, line 139: up until here, i think it is good. You simulate the prompt and put `n` to cancel the operation, nice job!

However, i would like you to change what follows (see my next comment)

cmd/kflex/ctx/delete_test.go
```diff
@@ -17,65 +17,136 @@ limitations under the License.
...
+
+	err := ExecuteCtxDelete(cp, ctxName, false)
+	wOut.Close()
+	var buf bytes.Buffer
+	buf.ReadFrom(rOut)
+	output := buf.String()
+	if !strings.Contains(output, "Warning: Context") || !strings.Contains(output, "Are you sure you want to delete this context?") {
```
rxinui, line 146: Although, this is correct, it is generally a bad idea to test a text-string. Reason is, kubeflex evolves and we might change the prompt content and forget to update it on the test side. Even if we do not forget, it is a pain to update the test as well.

Instead, I suggest to test the behaviour of cancelling the operation, which is deterministic. @rishi-jat 

CC: @pdettori

rxinui (changes requested):
Thank you, yet still some improvements to provide. Almost there, we appreciate your commitment.

rishi-jat:
@rxinui The test is now updated exactly as you requested:
It only checks that the context is not deleted if the user cancels, without checking any output text.

rxinui (commented):
/lgtm

@pdettori it's all yours now :)

rishi-jat:
Hi @pdettori,  
The latest changes are complete and have `/lgtm` from @rxinui.  
Could you please take a look and `/approve` if everything looks good?

rxinui:
@dumb0002 would you mind to run `/ok-to-test` :)

rishi-jat:
Hi @rxinui @dumb0002,
All tests are passing  if everything looks good, could you please add /approve? Thanks!

Issue #388, opened 2025-06-08
feature:


rxinui:
### Feature Description

> From @paolo
> 
> On the data corruption - we may not have to address it in this PR, but both Mike S. and I had issues in the past where that extension disappeared from the kubeconfig yaml. We suspected that some tools that write to the Kubeconfig did not preserve the extensions - did not investigate though which ones

### Proposed Solution

> Perhaps, we can push it to the creation of a command similar to `go vet` which analyses then solve configuration drift.
> 
> `kflex config diagnose` (identify drifts) or `kflex config tidy` (identify and autoresolve drifts)  to:
> 
> 1. Analyse a given kubeconfig
> 2. If missing kflex configuration (extensions), prompt or autoresolve the issue
> 3. If not possible to autoresolve, give a text-diagnosis to the end-user 

### Want to contribute?

- [ ] I would like to work on this issue.

### Additional Context

The definition of a command `config` enables flexibility to enhance kflex capabilities.

Along `tidy` and `diagnose`, we can imagine `set` as a subcommand to extend configuration of kubeconfig file from `kflex` CLI

pdettori:
I like the approach, this could be very useful especially as we add more info in the extensions for keeping track of relationships between control plane names and contexts.

rxinui:
Hi @rishi-jat thank you for your interest, however, this is assigned to @redpinecube who is actively working on it :)

PR #455, opened 2025-07-09
✨ feature : kflex config diagnose

redpinecube:
## Summary
Implement `kube config diagnose` to detect if the global Kubeflex extension is set.
Relates to #388


redpinecube:
@rxinui

PR #488, opened 2025-07-27
🐛 bug fix: Consolidate PostCreateHook error handling with retry logic in shared reconciler

### LINKED ISSUE #319: bug: No reaction to creation of PostCreateHook

RayyanSeliya:
## Summary

This update improves how the system handles situations when a PostCreateHook is missing. Instead of failing, it now waits and retries automatically without causing errors or interruptions. The retry logic is centralized, making the code cleaner and easier to maintain. This ensures that control planes referring to hooks that aren’t created yet will keep working smoothly until those hooks become available.

## Tested  e2e
Screenshot 2025-07-27 023632
also 
Screenshot 2025-07-27 023738

## Related issue(s)

Fixes #319

### DIFF
--- a/pkg/reconcilers/shared/postcreate_hook.go
+++ b/pkg/reconcilers/shared/postcreate_hook.go
@@ -34,11 +34,14 @@ import (
 
 	"github.com/kubestellar/kubeflex/api/v1alpha1"
 	"github.com/kubestellar/kubeflex/pkg/util"
+	"errors"
 )
 
 const (
 	FieldManager = "kubeflex"
 )
+
+var ErrPostCreateHookNotFound = errors.New("post create hook not found")
 
 type Vars struct {
 	Namespace        string
@@ -106,9 +109,7 @@ func (r *BaseReconciler) ReconcileUpdatePostCreateHook(ctx context.Context, hcp 
 		// Get hook definition
 		pch := &v1alpha1.PostCreateHook{}
 		if err := r.Client.Get(ctx, client.ObjectKey{Name: hookName}, pch); err != nil {
-			errs = append(errs, fmt.Errorf("failed to get PostCreateHook %s: %w", hookName, err))
-			allHooksApplied = false
-			continue
+			return fmt.Errorf("%w: %v", ErrPostCreateHookNotFound, err)
 		}
 
 		// Build variables with precedence: defaults -> global -> user vars -> system
--- a/pkg/reconcilers/shared/reconciler.go
+++ b/pkg/reconcilers/shared/reconciler.go
@@ -19,6 +19,7 @@ package shared
 import (
 	"context"
 	"strconv"
+	"time"
 
 	"github.com/pkg/errors"
 
@@ -66,6 +67,10 @@ func (r *BaseReconciler) UpdateStatusForSyncingError(hcp *tenancyv1alpha1.Contro
 	err := r.Status().Update(context.Background(), hcp)
 	if err != nil {
 		return ctrl.Result{}, errors.Wrap(e, err.Error())
+	}
+	if errors.Is(e, ErrPostCreateHookNotFound) {
+		// Requeue after 10 seconds, don't mark as failed
+		return ctrl.Result{Requeue: true, RequeueAfter: 10 * time.Second}, nil
 	}
 	return ctrl.Result{}, err
 }
--- a/test/e2e/test-postcreatehook-retry.sh
+++ b/test/e2e/test-postcreatehook-retry.sh
@@ -1,0 +1,66 @@
+#!/usr/bin/env bash
+
+set -e
+
+echo "🧹 Cleaning up any existing resources..."
+kubectl delete controlplane cp-missing-hook --ignore-not-found=true
+kubectl delete postcreatehook missing-hook --ignore-not-found=true
+
+echo ""
+echo "🔧 Creating ControlPlane referencing a missing PostCreateHook..."
+kubectl apply -f - <<EOF
+apiVersion: tenancy.kflex.kubestellar.org/v1alpha1
+kind: ControlPlane
+metadata:
+  name: cp-missing-hook
+spec:
+  backend: shared
+  postCreateHook: missing-hook
+  waitForPostCreateHooks: true
+  type: k8s
+EOF
+
+echo ""
+echo "⏳ Waiting 10s to check that ControlPlane is not marked as failed..."
+sleep 10
+
+echo ""
+echo "📋 ControlPlane status after 10s (should NOT be failed):"
+kubectl get controlplane cp-missing-hook -o jsonpath='{.status.conditions}' | jq '.'
+
+echo ""
+echo "🧪 Creating the missing PostCreateHook..."
+kubectl apply -f - <<EOF
+apiVersion: tenancy.kflex.kubestellar.org/v1alpha1
+kind: PostCreateHook
+metadata:
+  name: missing-hook
+spec:
+  templates:
+  - apiVersion: batch/v1
+    kind: Job
+    metadata:
+      name: job-missing-hook
+    spec:
+      template:
+        spec:
+          containers:
+          - name: demo
+            image: public.ecr.aws/docker/library/busybox:1.36
+            command: ["echo", "Hello from missing hook"]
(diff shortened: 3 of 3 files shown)


test/e2e/test-postcreatehook-retry.sh
```diff
@@ -0,0 +1,66 @@
...
+kubectl wait --for=condition=Ready controlplane/cp-missing-hook --timeout=90s
+
+echo ""
+echo "📊 FINAL STATUS:"
+kubectl get controlplane cp-missing-hook -o jsonpath='{.status}' | jq '.'
+
+echo ""
+echo "🧹 Cleaning up test resources..."
+kubectl delete controlplane cp-missing-hook --ignore-not-found=true
+kubectl delete postcreatehook missing-hook --ignore-not-found=true
```
rxinui, line 66: Can you please follow the codebase test flow? Having emojis are unnecessary and looks too ai generated.

Plain should be enough

rxinui: Thank you for pointing it out, will update the one with emojis later :)

RayyanSeliya: > Thank you for pointing it out, will update the one with emojis later :)

hey @rxinui should i update it  along with my commit ??would it be fine or i should inly focus on mine

RayyanSeliya: > @RayyanSeliya oh I see, I was not aware of this specific test. Could you follow this one instead https://github.com/kubestellar/kubeflex/blob/main/test/e2e/manage-type-external.sh
> 
> The one you mentioned was a bit out of context and necessary for kubestellar core project.

sure will follow this style and structure !!

rxinui: It would be better to do it on another issue, the refactoring of https://github.com/kubestellar/kubeflex/blob/main/test/e2e/test-postcreate-completion.sh

As it is written, it is mainly focus on applying CRD rather than using `kflex` (which is not a problem in itself but does not cover E2E as well...), will need to have a discussion with @pdettori around it. 

If an issue arise, i will post it

RayyanSeliya: > It would be better to do it on another issue, the refactoring of https://github.com/kubestellar/kubeflex/blob/main/test/e2e/test-postcreate-completion.sh
> 
> As it is written, it is mainly focus on applying CRD rather than using `kflex` (which is kinda a problem in itself but does not cover E2E as well...), will need to have a discussion with @pdettori around it.
> 
> If an issue arise, i will post it

sure thanks for the clarification @rxinui  do mention if this gets into a new issue will do it , and   any other issues you would recommend me to give a shot coz i want to strengthen my application for that unpaid intern of developing kubeflex doc site as i mentioned you on that post would love of any guidance ! i dont want to directly jump on issues without guidance coz it will affect the flow of  community i think so !

rxinui (commented):
/lgtm

@dumb0002 could you please `/ok-to-test`

pkg/reconcilers/shared/postcreate_hook.go
```diff
@@ -106,9 +109,7 @@ func (r *BaseReconciler) ReconcileUpdatePostCreateHook(ctx context.Context, hcp
 		// Get hook definition
 		pch := &v1alpha1.PostCreateHook{}
 		if err := r.Client.Get(ctx, client.ObjectKey{Name: hookName}, pch); err != nil {
-			errs = append(errs, fmt.Errorf("failed to get PostCreateHook %s: %w", hookName, err))
-			allHooksApplied = false
-			continue
+			return fmt.Errorf("%w: %v", ErrPostCreateHookNotFound, err)
```
MikeSpreitzer, line 112: FYI, https://pkg.go.dev/fmt@go1.24.5#Errorf says that you can use multiple `%w`.

pkg/reconcilers/shared/reconciler.go
```diff
@@ -67,6 +68,10 @@ func (r *BaseReconciler) UpdateStatusForSyncingError(hcp *tenancyv1alpha1.Contro
 	if err != nil {
 		return ctrl.Result{}, errors.Wrap(e, err.Error())
 	}
+	if errors.Is(e, ErrPostCreateHookNotFound) {
+		// Requeue after 10 seconds, don't mark as failed
```
MikeSpreitzer, line 72: But do please log the failure.
Can be done in follow-on.

test/e2e/test-postcreatehook-retry.sh
```diff
@@ -0,0 +1,107 @@
...
+  waitForPostCreateHooks: true
+  type: k8s
+EOF
+
+:
+: -------------------------------------------------------------------------
+: Verify ControlPlane is not marked as failed while waiting for hook
+:
+echo "Waiting 10s to check that ControlPlane is not marked as failed..."
+sleep 10
```
MikeSpreitzer, line 53: This 10 seconds is dead reckoning. We find that sometimes things are SUPRISINGLY slow in CI.

For this reason and others, it would be good if there were a positive indication of the expected state in the status of the ControlPlane. In `.status.conditions`, or `.status.erors`, or whatever.

test/e2e/test-postcreatehook-retry.sh
```diff
@@ -0,0 +1,107 @@
...
+
+echo "ControlPlane status after 10s (should NOT be failed):"
```
MikeSpreitzer, line 55: Being a test, shouldn't this fail if the `kubectl get` returns the unexpected result?

test/e2e/test-postcreatehook-retry.sh
```diff
@@ -0,0 +1,107 @@
...
+  - apiVersion: batch/v1
+    kind: Job
+    metadata:
+      name: job-missing-hook
+    spec:
+      template:
+        spec:
+          containers:
+          - name: demo
+            image: public.ecr.aws/docker/library/busybox:1.36
```
MikeSpreitzer, line 79: Do you really want to pin this to minor version 36?

test/e2e/test-postcreatehook-retry.sh
RayyanSeliya: You're absolutely right about the dead reckoning issue in CI. I've seen tests get flaky because of timing assumptions like this. Adding a positive status indication would be much better - maybe something like a "WaitingForPostCreateHooks" condition or adding hook status to the existing conditions.

test/e2e/test-postcreatehook-retry.sh
RayyanSeliya: Good catch , pinning to 1.36 is probably unnecessarily specific for a test. I can change that to just `busybox:latest` or `busybox:1` in the follow-on. Any preference on the tag strategy for test images?

RayyanSeliya:
hey @MikeSpreitzer   thx for the valuable feedback will get to u soon on the follow up pr i will be busy in some  university project work need to submit it will catch you for this after  2 3 days would it be fine ??

test/e2e/test-postcreatehook-retry.sh
```diff
@@ -0,0 +1,107 @@
...
+:
+echo "Cleaning up test resources..."
+kubectl delete controlplane cp-missing-hook --ignore-not-found=true
+kubectl delete postcreatehook missing-hook --ignore-not-found=true
+
+:
+: -------------------------------------------------------------------------
+: SUCCESS: Verified retry logic for missing PostCreateHook
+:
+echo "Test completed successfully"
```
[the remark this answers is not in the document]
greninja517: @RayyanSeliya , the red marker is generally due to the absence of blank line at the end of file. GitHub expects you to add blank line. It should fix the issue.

PR #493, opened 2025-08-04
✨feature : implement VerifyControlPlaneOnHostingCluster

redpinecube:
## Summary
Implements logic to check if a control plane is actually on a hosting cluster for CheckContextScopeKubeflexExtensionSet. 

## Related issue(s)
#388 
redpinecube/kubeflex#21

### DIFF
--- a/pkg/kubeconfig/extensions.go
+++ b/pkg/kubeconfig/extensions.go
@@ -19,6 +19,10 @@ package kubeconfig
 import (
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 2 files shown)


redpinecube:
@rxinui

redpinecube:
@rxinui

redpinecube:
@rxinui I didn’t use `IsAPIServerDeploymentReady` from `status_check` in this implementation. 
Do you think it’s important to include it? I can switch to using the controller-runtime client instead of the dynamic client if needed. Not sure which approach is preferred here.

rxinui (commented):
that seems good

/lgtm

@pdettori

PR #502, opened 2025-08-06
🐛 Fix Reconciliation logic to process PostCreateHooks only after Kubeconfig has been populated

### LINKED ISSUE #490: feature: Do not run PostCreateHooks until after the ControlPlane's in-cluster kubeconfig is written
### Feature Description

As seen in https://github.com/kubestellar/kubestellar/issues/2201 and https://github.com/kubestellar/kubestellar/issues/3098 , sometimes [the KubeStellar PCH for initializing an ITS] --- which is a Job --- has its containers launched before the in-cluster kubeconfig referenced in `ControlPlaneStatus.SecretRef` is populated. In particular, for ControlPlane type vcluster. This has long been a source of flakiness in KubeStellar's initialization of an ITS, see https://github.com/kubestellar/kubestellar/issues/2201 , https://github.com/kubestellar/kubestellar/issues/3098 , https://github.com/kubestellar/kubestellar/issues/2717 , and https://github.com/kubestellar/kubestellar/issues/3151 . This is also a contributing factor to https://github.com/kubestellar/kubestellar/issues/2932 .

### Proposed Solution

Do not execute the PostCreateHooks until after the Space is fully constructed.

### Want to contribute?

### Additional Context

_No response_

greninja517:
## Summary
This PR updates and reorders the main reconciliation logic to process PCHs only after kubeconfig is populated in the secret referenced by `ControlPlaneStatus.SecretRef` and infrastructure is ready for each type of reconcilers. Also, `vcluster` reconciler logic is accurately reordered. The redundant `switch` blocks in the main controller reconciler is also addressed.

## Related issue(s)

Fixes #490


internal/controller/controlplane_controller.go
```diff
@@ -145,85 +145,65 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
-	}
-
-	// ALWAYS call the type-specific reconciler to handle infrastructure (API server, etc.)
-	log.Info("Calling type-specific reconciler for infrastructure setup", "type", hcp.Spec.Type)
+	// PHASE 1: Type-specific controlplane reconciliation
+	log.Info("Phase 1: Infrastructure setup and kubeconfig creation", "type", hcp.Spec.Type)
 	var reconcileResult ctrl.Result
 	var reconcileError error
 
+	var reconciler interface {
```
rxinui, line 153: This is a bad idea and generally not a good practice to define an interface within a function.

In addition, there will be a `Reconciler` interface, which is currently implemented within `lxf/main`. See https://github.com/kubestellar/kubeflex/blob/8d144ee0f9bd1457e952412339719b280dd54e0c/pkg/reconcilers/shared/reconciler.go#L47

internal/controller/controlplane_controller.go
```diff
@@ -145,85 +145,65 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+		Reconcile(ctx context.Context, hcp *tenancyv1alpha1.ControlPlane) (ctrl.Result, error)
+		ReconcileUpdatePostCreateHook(ctx context.Context, hcp *tenancyv1alpha1.ControlPlane) error
```
rxinui, line 155: This should be defined within a separate interface.

As shown above within `lxf/main` branch, an interface is created for ControlPlane object and its reconciling.

Since PostCreateHook is a separate kubeflex Object, the same logic should be applied --> separate reconciler interface for it.

internal/controller/controlplane_controller.go
```diff
@@ -145,85 +145,65 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 
-	// NOW determine overall CP readiness based on BOTH API server AND hooks
+	// PHASE 2: PostCreateHook processing
+	if hcp.Spec.PostCreateHook != nil || len(hcp.Spec.PostCreateHooks) > 0 {
+		log.Info("Processing PostCreateHooks with complete kubeconfig")
+
+		if err := reconciler.ReconcileUpdatePostCreateHook(ctx, hcp); err != nil {
+			log.Error(err, "Failed to process PostCreateHooks")
+			// Don't return error immediately - let status logic handle it
+		}
```
rxinui, line 198: See above comment regarding interface.

pkg/reconcilers/vcluster/reconciler.go
```diff
@@ -142,10 +142,19 @@ func (r *VClusterReconciler) Reconcile(ctx context.Context, hcp *tenancyv1alpha1
...
+		}
+		return r.UpdateStatusForSyncingError(hcp, err)
+	}
+
+	// Move this call after ReconcileKubeconfigSecret()
 	r.UpdateStatusWithSecretRef(hcp, util.VClusterKubeConfigSecret,
 		util.KubeconfigSecretKeyVCluster, util.KubeconfigSecretKeyVClusterInCluster)
 
-	if hcp.Spec.PostCreateHook != nil &&
+	if len(hcp.Spec.PostCreateHooks) > 0 &&
```
rxinui, line 157: What happened to `hcp.Spec.PostCreateHook` ?

rxinui (changes requested):
Some inconsistency to correct.

@MikeSpreitzer you should have a look as well when time allows you.

internal/controller/controlplane_controller.go
```diff
@@ -145,85 +145,65 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 	case tenancyv1alpha1.ControlPlaneTypeExternal:
-		reconciler := external.New(r.Client, r.Scheme, r.Version, r.ClientSet, r.DynamicClient, r.EventRecorder)
-		reconcileResult, reconcileError = reconciler.Reconcile(ctx, hcp)
+		reconciler = external.New(r.Client, r.Scheme, r.Version, r.ClientSet, r.DynamicClient, r.EventRecorder)
 	default:
 		reconcileError = fmt.Errorf("unsupported control plane type: %s", hcp.Spec.Type)
 	}
 
-	// If type-specific reconciler failed, return the error
 	if reconcileError != nil {
```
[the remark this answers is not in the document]
greninja517: It is defined at line 151.

pkg/reconcilers/vcluster/reconciler.go
greninja517: I removed `hcp.Spec.PostCreateHook` as it was depreciated, and plural form can also handle the same check. But, as for now, should i reintroduce it just for backward compatibility ? Later, we will update `hcp.Spec.PostCreateHook` across the project at once.

internal/controller/controlplane_controller.go
greninja517: Okay. Since this interface is not exported and used by any other files. So, should I define it in the same file outside the function ? Or, define it in the shared reconciler at `pkg/reconcilers/shared/reconciler.go` ? Is the interface name okay ?

internal/controller/controlplane_controller.go
```diff
@@ -145,85 +145,65 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+
+	// Call the type-specific reconciler to handle infrastructure and kubeconfig setup
+	reconcileResult, reconcileError = reconciler.Reconcile(ctx, hcp)
+	if reconcileError != nil {
+		log.Error(reconcileError, "Type-specific reconciliation failed")
 		return reconcileResult, reconcileError
 	}
 
-	// Refresh the hcp object after infrastructure reconciliation
+	// Refresh hcp object after infrastructure reconciliation
```
MikeSpreitzer, line 185: This is not an improvement in English grammar.

pkg/reconcilers/vcluster/reconciler.go
MikeSpreitzer: This whole `if ... { ... }` statement can and should be deleted, it is redundant with the handling in `
func (r *ControlPlaneReconciler) Reconcile`. (And the fact that it does only a subset of what the common one does is why this is not a functional error.)

internal/controller/controlplane_controller.go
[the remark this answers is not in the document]
greninja517: Ok, noted.

pkg/reconcilers/vcluster/reconciler.go
MikeSpreitzer: Yes, remove the redundancy. Just keep the one common call from `func (r *ControlPlaneReconciler) Reconcile`.

internal/controller/controlplane_controller.go
greninja517: Okay, will implement compile check for each implementing control plane type in a follow on PR.

greninja517:
@MikeSpreitzer All requested changes are done, except this one https://github.com/kubestellar/kubeflex/pull/502#discussion_r2263203043 , which will done in a follow-on PR.

internal/controller/controlplane_controller.go
```diff
@@ -145,85 +145,65 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 	if err := r.Get(ctx, client.ObjectKey{Name: hcp.Name}, hcp); err != nil {
 		log.Error(err, "Failed to refresh ControlPlane after infrastructure reconciliation")
 		return ctrl.Result{}, err
 	}
 
-	// NOW determine overall CP readiness based on BOTH API server AND hooks
+	// PHASE 2: PostCreateHook processing
```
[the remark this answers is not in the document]
MikeSpreitzer: Oops, sorry, I used that ambiguous word "should". I meant it in the sense of "this would be true in an ideal world, but is not true in the real world now".

internal/controller/controlplane_controller.go
```diff
@@ -145,75 +146,32 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
-			}
-		}
-	}
-
-	// ALWAYS call the type-specific reconciler to handle infrastructure (API server, etc.)
-	log.Info("Calling type-specific reconciler for infrastructure setup", "type", hcp.Spec.Type)
+	// PHASE 1: Type-specific controlplane reconciliation
+	log.Info("Phase 1: Infrastructure setup and kubeconfig creation", "type", hcp.Spec.Type)
 	var reconcileResult ctrl.Result
 	var reconcileError error
```
MikeSpreitzer, line 152: FYI, these variable declarations could be simply deleted if the one statement that assigns to these variables used `:=` rather than `=`.

pkg/reconcilers/vcluster/reconciler.go
```diff
@@ -70,8 +70,8 @@ func (r *VClusterReconciler) Reconcile(ctx context.Context, hcp *tenancyv1alpha1
 
 	if err := r.BaseReconciler.ReconcileNamespace(ctx, hcp); err != nil {
 		if util.IsTransientError(err) {
-            return ctrl.Result{}, err
-        }
+			return ctrl.Result{}, err
+		}
```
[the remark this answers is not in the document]
MikeSpreitzer: I still see spaces being replaced by tabs here.

pkg/reconcilers/vcluster/reconciler.go
```diff
@@ -142,28 +142,17 @@ func (r *VClusterReconciler) Reconcile(ctx context.Context, hcp *tenancyv1alpha1
...
-			if util.IsTransientError(err) {
-				return ctrl.Result{}, err
-			}
-			return r.UpdateStatusForSyncingError(hcp, err)
+	// try to update kubeconfig secret to add the incluster config unconditionally
+	if err := r.ReconcileKubeconfigSecret(ctx, hcp); err != nil {
+		if util.IsTransientError(err) {
+			return ctrl.Result{}, err
 		}
+		return r.UpdateStatusForSyncingError(hcp, err)
```
MikeSpreitzer, line 150: What if this failed because some setup is not done yet, but will be done in a little while? In this case it is not right to give up.

pkg/reconcilers/shared/reconciler.go
```diff
@@ -41,6 +41,12 @@ const (
 	FieldOwner = "kubeflex.kubestellar.io"
 )
 
+// Implemented by all controlplane types for central PCH processing
+type ControlPlaneTypeReconciler interface {
+	Reconcile(context.Context, *tenancyv1alpha1.ControlPlane) (ctrl.Result, error)
```
MikeSpreitzer, line 46: Why is this different from https://github.com/kubestellar/kubeflex/blob/lxf/main/pkg/reconcilers/shared/reconciler.go#L43-L51 ?

pkg/reconcilers/vcluster/reconciler.go
greninja517: Next reconciliation loop will handle even if `r.ReconcileKubeconfigSecret(..)` fails. The main controller reconciliation will also be stopped so PCHs won't be processed in that loop. I think it's fine.
Adding retries will block the reconciliation, which I think is not that much ideal. What else we can do for this ?

pkg/reconcilers/shared/reconciler.go
greninja517: https://github.com/kubestellar/kubeflex/pull/502#discussion_r2257968909; here it was suggested to have a different interface for only PCH related stuff. But, we can make it same as that of `lxf/main` for consistency. Should I go for it ?

pkg/reconcilers/vcluster/reconciler.go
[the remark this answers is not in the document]
greninja517: What I see when I used `cat -A <file_name>` command is:
In `main` branch:
```shell
^Iif err := r.BaseReconciler.ReconcileNamespace(ctx, hcp); err != nil {$
^I^Iif util.IsTransientError(err) {$
            return ctrl.Result{}, err$
        }$
^I^Ireturn r.UpdateStatusForSyncingError(hcp, err)$
^I}$
```

In my `feature` branch:
```shell
^Iif err := r.BaseReconciler.ReconcileNamespace(ctx, hcp); err != nil {$
^I^Iif util.IsTransientError(err) {$
^I^I^Ireturn ctrl.Result{}, err$
^I^I}$
^I^Ireturn r.UpdateStatusForSyncingError(hcp, err)$
^I}$
```
Originally, tabs and spaces both were used inconistently in that part of file. But, for consistency; we should use tabs as it is the default Go convention. Also, rest of the parts of file has tabs everywhere as well.

pkg/reconcilers/vcluster/reconciler.go
MikeSpreitzer: `r.UpdateStatusForSyncingError` is about updating the ControlPlane's status. It does _not_ normally cause a requeue.

internal/controller/controlplane_controller.go
```diff
@@ -223,17 +179,41 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+		return ctrl.Result{RequeueAfter: time.Second * 15}, nil
+	}
+
+	// PHASE 3: PostCreateHook processing
+	if hcp.Spec.PostCreateHook != nil || len(hcp.Spec.PostCreateHooks) > 0 {
+		log.Info("Processing PostCreateHooks with complete kubeconfig")
+
+		if err := reconciler.ReconcileUpdatePostCreateHook(ctx, hcp); err != nil {
+			log.Error(err, "Failed to process PostCreateHooks")
+			// Don't return error immediately - let status logic handle it
```
MikeSpreitzer, line 206: The error value evaporates at the close bracket on the next line, so no logic will handle it.

MikeSpreitzer:
It is worth noting that `test/e2e/run.sh` does NOT exercise the PCH functionality of a vcluster type of ControlPlane.

MikeSpreitzer:
BTW, @greninja517 and @rxinui: we seem to be converging on having two different types with the same name (ControlPlaneReconciler). This works because the two are defined in different namespaces. Even so, this is not the best way to go. I recommend that in follow-on PRs we adjust the convergence so that the two different types have two different names.

pkg/reconcilers/shared/reconciler.go
```diff
@@ -72,7 +78,7 @@ func (r *BaseReconciler) UpdateStatusForSyncingError(hcp *tenancyv1alpha1.Contro
 		// Requeue after 10 seconds, don't mark as failed
 		return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
 	}
-	return ctrl.Result{}, err
+	return ctrl.Result{}, e
```
MikeSpreitzer, line 81: This change is relevant to all the calls on this func, and there are many. I think that this change is an improvement, but I suspect that @pdettori was thinking differently.
/cc @pdettori

greninja517:
> > It is worth noting that `test/e2e/run.sh` does NOT exercise the PCH functionality of a vcluster type of ControlPlane. I do not expect this PR to fix the e2e testing. But any PR should have some sort of evidence of successful testing.
> 
> It would be nice that this PR comes wit ha E2E test

What about parameterizing these scripts test/e2e/test-postcreate-completion.sh and test/e2e/test-postcreatehook-retry.sh with controlplaneTypes and invoking them through test/e2e/run.sh for both k8s and vcluster type.

rxinui:
I'd like the following PR https://github.com/kubestellar/kubeflex/pull/520 to be merged before this one.

test/e2e/test-postcreate-completion.sh
```diff
@@ -1,11 +1,14 @@
 #!/usr/bin/env bash
 
-echo "🧪 Creating PostCreateHook..."
-kubectl apply -f - <<EOF
+CP_TYPE=${1:-k8s}
+echo "Testing PostCreateHook completion behavior with ${CP_TYPE} control plane..."
+echo ""
+echo "Creating PostCreateHook (${CP_TYPE})..."
```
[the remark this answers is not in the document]
greninja517: Should I remove everywhere or add it here ?

test/e2e/test-kubeconfig-access.sh
```diff
@@ -0,0 +1,132 @@
...
+    ;;
+esac
+
+echo "Using secret: $SECRET_NAME with key: $SECRET_KEY for $CP_TYPE"
+
+:
+: -------------------------------------------------------------------------
+: Clean up any existing resources
+:
+kubectl --context kind-kubeflex delete controlplane ${CP_NAME} --ignore-not-found=true
```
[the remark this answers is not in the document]
greninja517: Not, Really. We have to specify the context here (and everywhere) in the scripts because `kflex create ....` switches the context to controlplane context. So, it's not guaranteed that we are always interacting with the kubeflex hosting cluster.

test/e2e/manage-type-k8s.sh
```diff
@@ -77,6 +77,12 @@ kubectl --context kind-kubeflex wait --for=condition=Established crd cr1s.synthe
 kubectl --context cp1 create ns e2e
 kubectl --context cp1 wait --for=jsonpath='{.status.phase}'=Active ns/e2e --timeout=120s
 
+:
+: -------------------------------------------------------------------------
+: Test PostCreateHook kubeconfig access
+:
+${SRC_DIR}/test-kubeconfig-access.sh -t k8s
+
```
MikeSpreitzer, line 85: Since this test does not involve `cp1`, it could come after the deletion of `cp1`. That would be better, less resource stress.

internal/controller/controlplane_controller.go
```diff
@@ -223,17 +179,42 @@ func (r *ControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 
-	// NOW determine overall CP readiness based on BOTH API server AND hooks
-	log.Info("Determining final control plane readiness")
-
-	// Check API server readiness
+	// PHASE 2: Check API server readiness
 	apiServerReady, err := util.IsAPIServerDeploymentReady(log, r.Client, *hcp)
 	if err != nil {
 		log.Error(err, "Error checking API server readiness", "controlPlane", hcp.Name)
 	}
```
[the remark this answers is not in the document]
greninja517: Revised.

PR #504, opened 2025-08-09
🌱 Bump the kubectl image from 1.30.12 to 1.30.14

MikeSpreitzer:
## Summary
Bump kubectl image to pick up security fixes.

## Related issue(s)

Fixes #

### DIFF
--- a/chart/templates/install-hooks.yaml
+++ b/chart/templates/install-hooks.yaml
@@ -9,7 +9,7 @@ spec:
     spec:
       containers:
       - name: "{{ .Release.Name }}"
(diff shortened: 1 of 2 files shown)


francostellari:
@MikeSpreitzer or @Rupam-It , can we also update the value in the various charts in all repos?

MikeSpreitzer:
Perhaps it would be helpful to make the kubectl image version a paramter of the KubeFlex Helm chart too. In a way that plays nice with its inclusion into the ks/ks core Helm chart, of course.

PR #515, opened 2025-08-14
🐛 ingress name and clusterrolebinding is attached to controlplane name

### LINKED ISSUE #510: task: add k3s token as secret
When `k3s server` starts, it generates a k3s node token allowing worker node to join the cluster.

This k3s token stores a default value as a file available by default at `/var/lib/rancher/k3s/server/node-token` (or can be override by `k3s server --data-dir=newpath`)

Store this token as part of k3s secret `k3s-config` with secret key `token` and displays it on terminal.

rxinui:
## Summary

Before this PR, it was not possible to create 2 controlplanes of type k3s as they were sharing the common name for ingress and clusterrolebinding, putting the reconciling loop in error.

Now ingress and clusterrolebinding leverage controlplane name to their resource name.

## Related issue(s)

Fixes #510

### DIFF
--- a/pkg/reconcilers/k3s/apiserver.go
+++ b/pkg/reconcilers/k3s/apiserver.go
@@ -69,8 +69,8 @@ func containerImage() string {
 }
 
 // serverTLSSAN returns the TLS SAN value expected by k3s server command
-func serverTLSSAN(cfg *shared.SharedConfig) string {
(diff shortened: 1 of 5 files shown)


rxinui:
@dumb0002 FYI, bringing fix to run 2 or more controlplane of k3s at the same time

PR #521, opened 2025-08-19
add cluster param to init command

antedotee:
## Add cluster name parameter to init command

This PR adds support for specifying a custom cluster name when creating a kind cluster with `kflex init.

### Changes
- Added `--cluster-name` flag to the `init` command (defaults to "kubeflex" for backward compatibility)
- Modified `CreateKindCluster()` and `checkKindInstanceExists()` functions to accept and use the cluster name parameter
- Updated cluster creation logic to use the provided name instead of hardcoded "kubeflex"

### Usage
```bash
# Create cluster with custom name
kflex init --create-kind --cluster-name my-test-cluster

# Creates kind cluster named "my-test-cluster" with context "kind-my-test-cluster"
```

### Motivation
Previously, users could only create one kubeflex installation at a time since the cluster name was hardcoded to "kubeflex". This prevented testing multiple configurations or features in parallel. Now users can create multiple isolated kubeflex environments with different cluster names.

Fixes the issue where parallel kubeflex installations would conflict due to identical cluster names.

### DIFF
--- a/cmd/kflex/init/init.go
+++ b/cmd/kflex/init/init.go
@@ -57,6 +57,7 @@ func Command() *cobra.Command {
 			domain, _ := flagset.GetString(DomainFlag)
 			externalPort, _ := flagset.GetInt(ExternalPortFlag)
 			hostContainer, _ := flagset.GetString(HostContainerNameFlag)
+			clusterName, _ := flagset.GetString("cluster-name")
 			done := make(chan bool)
 			var wg sync.WaitGroup
 			var isOCP bool
@@ -76,7 +77,7 @@ func Command() *cobra.Command {
 				if isOCP {
 					return fmt.Errorf("openShift cluster detected on existing context\nSwitch to a non-OpenShift context with `kubectl config use-context <context-name>` and retry")
 				}
-				cluster.CreateKindCluster(chattyStatus)
+				cluster.CreateKindCluster(chattyStatus, clusterName)
 			}
 
 			cp := common.NewCP(kubeconfig)
@@ -90,6 +91,7 @@ func Command() *cobra.Command {
 	flagset.StringP(DomainFlag, "d", "localtest.me", "domain for FQDN")
 	flagset.StringP(HostContainerNameFlag, "n", "kubeflex-control-plane", "Name of the hosting cluster container (kind or k3d only)")
 	flagset.IntP(ExternalPortFlag, "p", 9443, "external port used by ingress")
+	flagset.String("cluster-name", "kubeflex", "Name of the kind cluster to create")
 	return command
 }
 
--- a/cmd/kflex/init/cluster/kind.go
+++ b/cmd/kflex/init/cluster/kind.go
@@ -57,7 +57,7 @@ func installKind() error {
 	return nil
 }
 
-func checkKindInstanceExists() (bool, error) {
+func checkKindInstanceExists(clusterName string) (bool, error) {
 	cmd := exec.Command("kind", "get", "clusters")
 	out, err := cmd.Output()
 	if err != nil {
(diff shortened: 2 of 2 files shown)


cmd/kflex/init/init.go
```diff
@@ -57,7 +57,12 @@ func Command() *cobra.Command {
 			domain, _ := flagset.GetString(DomainFlag)
 			externalPort, _ := flagset.GetInt(ExternalPortFlag)
 			hostContainer, _ := flagset.GetString(HostContainerNameFlag)
-			clusterName, _ := flagset.GetString("cluster-name")
+			
+			// Handle positional cluster name parameter
+			clusterName := "kind-kubeflex" // default
```
rxinui, line 62: It is better thank you. Yet, I'd export that as a constant and refer the constant here.

rxinui (changes requested):
Thank you for the change. There is a small improvement to do.

On top of it, it is really important to include E2E test. There is one available probably for `kflex init -c` but there is none for `kflex init -c my-cluster-name`.

This is crucial.

pdettori:
@antedotee could you please resolve the conflicts and update the PR?

test/e2e/run.sh
```diff
@@ -37,6 +37,7 @@ ${SRC_DIR}/manage-type-k8s.sh
 ${SRC_DIR}/manage-type-vcluster.sh
 ${SRC_DIR}/manage-type-external.sh
 ${SRC_DIR}/manage-ctx.sh
+${SRC_DIR}/test-custom-cluster-name.sh
```
pdettori, line 40: Please move this to the very end after the test-post create

test/e2e/test-custom-cluster-name.sh
```diff
@@ -0,0 +1,70 @@
...
+set -e # exit on error
+
+CUSTOM_CLUSTER_NAME="test-custom-cluster"
+
+:
+: -------------------------------------------------------------------------
+: "Create kubeflex cluster with custom name"
+:
+if [ ! -z "$(bin/kflex init -c "${CUSTOM_CLUSTER_NAME}" 2>&1)" ]; then
+    echo "ERROR: Failed to create kubeflex cluster with custom name"
```
pdettori, line 29: Since the previous kubeflex hosting cluster still exists, this cluster conflicts on port 9080 - the CI e2e test shows:

```
Bind for 0.0.0.0:9080 failed: port is already allocated
```

Before trying to create the cluster with the new name it should use the kind command to delete the existing kind cluster so there is no conflict on that port.

pdettori: See this run with e2E test failing because of ports conflicts: https://github.com/kubestellar/kubeflex/actions/runs/17498172259/job/49802603324?pr=521

MikeSpreitzer:
Ping @antedotee

test/e2e/test-custom-cluster-name.sh
```diff
@@ -0,0 +1,57 @@
...
+    exit 1
+fi
+
+echo "✓ kflex init command argument parsing works correctly"
+
+:
+: -------------------------------------------------------------------------
+: "SUCCESS: Custom cluster name test completed"
+:
+echo "✓ Custom cluster name functionality validation passed"
```
MikeSpreitzer: Not loving the red circle-dash here.
The last line should end with a linebreak (newline (AKA LF), if this is in Linux/Unix format).

Image

cmd/kflex/init/init.go
MikeSpreitzer: There is a symbolic constant for this now.

cmd/kflex/init/init_test.go
MikeSpreitzer: These are checksum tests of two tiny details. I prefer behavioral tests.

PR #480, opened 2025-07-18
✨ feat: add generated typed clientset for tenancy/v1alpha1

rishi-jat:
📝 Description

This PR generates and adds a typed clientset for the tenancy/v1alpha1 custom resources under pkg/generated/clientset/ using the Kubernetes client-gen tool.

✅ No hand-written code was modified; only generated files were added.

📌 Purpose

This enables type-safe and efficient access to the tenancy.kflex.kubestellar.org CRDs for any code using Kubernetes client-go interfaces, including controllers and CLI tools.

🔧 Implementation Details
	•	Generated using [k8s.io/code-generator](https://github.com/kubernetes/code-generator)
	•	Supports the ControlPlane resource under the internal versioned clientset
	•	Includes fake clients for unit testing and clientset interfaces

🔗 Fixes

Resolves #277

Issue #573, opened 2025-11-07
bug: wrong package path for generated typed/external stubs


MikeSpreitzer:
### Describe the bug

There was a goof-up in PR #480. Look at the package path `pkg/generated/clientset/versioned/typed/v1alpha1/internalversion`. That last part, `/internalversion` does not belong there.

In Kubernetes, there is a dichotomy between "internal" AKA "untyped" stubs and "external" AKA "typed" stubs. In internal/untyped ones are only used by developers of resources that are built into apiservers. Ordinary clients use the external/typed stubs.

### Steps To Reproduce

This came from code examination, not testing.

### Expected Behavior

Following the usual patterns.

Issue #277, opened 2024-08-14
feature: Add typed clients


MikeSpreitzer:
### Feature Description

Generate typed client code (clientset, informers and listers), to support clients that use plain Kubernetes. Currently clients have to use controller-runtime's `client` or the plain Kubernetes "dynamic"  client.

### Proposed Solution

Do it.

### Want to contribute?

- [ ] I would like to work on this issue.

### Additional Context

https://github.com/kubestellar/kubestellar/pull/2394

rxinui:
In addition to #308, i can also look at that.

RONAK-AI647:
may i work on this issue , if it still persists!!!

MikeSpreitzer:
I see no informers.

MikeSpreitzer:
PR #480 is also deficient in this way: https://github.com/kubestellar/kubeflex/pull/480#issuecomment-3618458617

PR #580, opened 2025-11-19
🐛fix: correct typed client package path

rishi-jat:
## Summary
Fixes the incorrect package structure introduced in #480. The typed clientset 
was generated under `v1alpha1/internalversion/` but should be directly under 
`v1alpha1/` since these are external/typed clients, not internal ones.

Changes:
- Moved files from `typed/v1alpha1/internalversion/` to `typed/v1alpha1/`
- Updated package names and import paths accordingly
- Removed the internalversion directory

Fixes #480.


MikeSpreitzer:
> Fixes https://github.com/kubestellar/kubeflex/pull/480.

Typo here. I suspect you meant to reference Issue #573

MikeSpreitzer (approved):
This is an improvement.
/lgtm
But problems still remain. See recent comments on #277 and #480

PR #587, opened 2025-12-12
🐛 remove the +required marker for backend

### LINKED ISSUE #586: bug: adoption creates invalid ControlPlane object
### Describe the bug

https://github.com/kubestellar/kubeflex/blob/fb35e66a6bba6f27ac9a462909ee4510feaf8743/api/v1alpha1/controlplane_types.go#L29 says that `.spec.backend` is required. The type of this field is an enumerated type with just two valid values: "shared" and "dedicated". In Go there is no such thing as a struct with one of its defined fields absent; the only question is whether the field's value is its type's zero value or not.

https://github.com/kubestellar/kubeflex/blob/fb35e66a6bba6f27ac9a462909ee4510feaf8743/cmd/kflex/adopt/adopt.go#L128 creates a ControlPlane whose `.Spec.Backend` is the empty string.

### Steps To Reproduce

In `test/e2e/run.sh`, interrupt it while `manage-type-external.sh` has the external ControlPlane in existence; see that it does indeed break the requirement that `.spec.backend` exists (i.e., is not the empty string).

```console
mspreitz@mjs13 kubeflex % kubectl --context kind-kubeflex get cps ext -o yaml
apiVersion: tenancy.kflex.kubestellar.org/v1alpha1
kind: ControlPlane
metadata:
  creationTimestamp: "2025-12-11T06:15:44Z"
  finalizers:
  - kflex.kubestellar.org/finalizer
  generation: 1
  name: ext
  resourceVersion: "2323"
  uid: 768e63cc-b590-4f94-a6b8-b190edd53fda
spec:
  bootstrapSecretRef:
    inClusterKey: kubeconfig-incluster
    name: ext-bootstrap
    namespace: kubeflex-system
  postCreateHook: list-controller
  tokenExpirationSeconds: 31536000
  type: external
  waitForPostCreateHooks: false
status:
  conditions:
  - lastTransitionTime: "2025-12-11T06:15:44Z"
    lastUpdateTime: "2025-12-11T06:15:44Z"
    message: ""
    reason: ReconcileSuccess
    status: "True"
    type: Synced
  - lastTransitionTime: "2025-12-11T06:15:44Z"
    lastUpdateTime: "2025-12-11T06:15:44Z"
    message: ""
    reason: Available
    status: "True"
    type: Ready
  observedGeneration: 1
  postCreateHookCompleted: true
  postCreateHooks:
    list-controller: true
  secretRef:
    inClusterKey: kubeconfig-incluster
    key: kubeconfig
    name: admin-kubeconfig
    namespace: ext-system
```

### Expected Behavior

The code is self-consistent.

### Additional Context

This inconsistency was discovered in CI of #583 . Why was this not discovered earlier?

pdettori:
## Summary

relax the "required" condition to make .spec.backend optional.

## Related issue(s)

Fixes #586

### DIFF
--- a/api/v1alpha1/controlplane_types.go
+++ b/api/v1alpha1/controlplane_types.go
@@ -26,7 +26,10 @@ type ControlPlaneSpec struct {
 	// +required
 	Type ControlPlaneType `json:"type,omitempty"`
 	// backend refers to the database type used by the control plane
-	// +required
+	// For external control planes this field is optional; controller
+	// logic can ignore it for `external` type instances. Removed
+	// the `+required` marker so the CRD validation doesn't force a
+	// value for adopt flows that don't set a backend.
 	Backend BackendDBType `json:"backend,omitempty"`
 	// bootstrapSecretRef contains a reference to the kubeconfig used to bootstrap adoption of
 	// an external cluster
--- a/config/crd/bases/tenancy.kflex.kubestellar.org_controlplanes.yaml
+++ b/config/crd/bases/tenancy.kflex.kubestellar.org_controlplanes.yaml
@@ -56,8 +56,12 @@ spec:
             description: ControlPlaneSpec defines the desired state of ControlPlane
             properties:
               backend:
-                description: backend refers to the database type used by the control
-                  plane
+                description: |-
+                  backend refers to the database type used by the control plane
+                  For external control planes this field is optional; controller
+                  logic can ignore it for `external` type instances. Removed
+                  the `+required` marker so the CRD validation doesn't force a
+                  value for adopt flows that don't set a backend.
                 enum:
                 - shared
                 - dedicated


api/v1alpha1/controlplane_types.go
```diff
@@ -26,7 +26,10 @@ type ControlPlaneSpec struct {
 	// +required
 	Type ControlPlaneType `json:"type,omitempty"`
 	// backend refers to the database type used by the control plane
```
MikeSpreitzer: Need a period at the end of this sentence.

api/v1alpha1/controlplane_types.go
```diff
@@ -26,7 +26,10 @@ type ControlPlaneSpec struct {
...
-	// +required
+	// For external control planes this field is optional; controller
+	// logic can ignore it for `external` type instances. Removed
```
MikeSpreitzer: If I understand correctly, the following would be a simpler clearer statement.

> For external control planes this field is irrelevant and may be omitted (I.e., contain the empty string).

MikeSpreitzer (approved):
This is good enough to merge.
/lgtm
/approve

Issue #588, opened 2025-12-15
bug: Latest Helm chart has bogus controller-manager image reference


MikeSpreitzer:
### Describe the bug

```shell
helm template test oci://ghcr.io/kubestellar/kubeflex/chart/kubeflex-operator --version v0.9.2 > /tmp/kf.yaml
```

Then look at the definition of the `kubeflex-controller-manager` Deployment; you will find that the `manager` container says

```yaml
        image: ghcr.io/kubestellar/kubeflex/manager:v0.9.2
```

But there is no such image; the correct tag lacks the initial `v`:

```console
mspreitz@mjs13 e2e % docker buildx imagetools inspect ghcr.io/kubestellar/kubeflex/manager:v0.9.2
ERROR: ghcr.io/kubestellar/kubeflex/manager:v0.9.2: not found

mspreitz@mjs13 e2e % docker buildx imagetools inspect ghcr.io/kubestellar/kubeflex/manager:0.9.2 
Name:      ghcr.io/kubestellar/kubeflex/manager:0.9.2
MediaType: application/vnd.oci.image.index.v1+json
Digest:    sha256:6482bafdd309179f4a3d7fa349bd4cfbab862e3c82b90a2c662bbc17d4fe21b5
           
Manifests: 
  Name:      ghcr.io/kubestellar/kubeflex/manager:0.9.2@sha256:bddb5af7291192506641844298f1b7ac23fe9637875e28e43dcaadcbc98b1fc9
  MediaType: application/vnd.oci.image.manifest.v1+json
  Platform:  linux/amd64
             
  Name:      ghcr.io/kubestellar/kubeflex/manager:0.9.2@sha256:b4a48e37b32a64ef7ef32c7ad97b89a09862d518fc77f331500606a881ed1eca
  MediaType: application/vnd.oci.image.manifest.v1+json
  Platform:  linux/arm64
```

In earlier versions of the Helm chart, the image tag was correct.

### Steps To Reproduce

Try to use the chart.

### Expected Behavior

The chart references images that exist.

### Additional Context

This causes https://github.com/kubestellar/kubestellar/pull/3453 to fail testing.

shivansh-source:
update this bug causes prblem for [kubestellar/kubestellar#3574](url)

shivansh-source:
causing problem in [kubestellar/kubestellar#3574](https://github.com/kubestellar/kubestellar/pull/3574)

PR #519, opened 2025-08-16
🐛 fix: resolve controller manager image update issue 

RayyanSeliya:
## Summary

This PR fixes issue #453 where the `kubeflex-controller-manager` pod was not updating with new images after running `make install-local-chart`. The root cause was a problematic `sed` command in the Makefile's `chart` target that caused image references to remain as `<placeholder>` instead of the actual image name.

## Related issue(s)

Fixes #453

## Problem

After running `make install-local-chart`, the kubeflex-controller-manager pod was not restarting with the new image, continuing to run the old version. The deployment object was not being updated with the new image reference, causing the pod to continue running the old version.

**Evidence from the original issue:**
- New image was built: `ko.local/manager:f72a332`
- Helm upgrade completed successfully
- But pod name remained the same: `kubeflex-controller-manager-699d869487-p6wsx`
- No new pod was created - indicating the Deployment spec didn't change

## Root Cause

The issue was in the Makefile's `chart` target:
```makefile
# BROKEN - This sed command was failing silently
cd config/manager && $(KUSTOMIZE) edit set image controller=$(shell echo ${IMG} | sed 's/\(:.*\)v/\1/')
```

The `sed` command was designed to remove 'v' from version tags, but `IMAGE_TAG` uses git commit hashes (like `f72a332`), not version tags with 'v'. This caused the image reference to remain as `<placeholder>` in the generated chart.

### 1. Fixed Makefile chart target
```makefile
# FIXED - Simplified to use the full image reference
cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
```

### 2. Enhanced Deployment configuration
- Added explicit `RollingUpdate` strategy for better update control

### 3. Added comprehensive E2E test
- Created `test/e2e/test-controller-image-update.sh` to verify the fix
- Tests image reference updates, pod recreation, and deployment stability
- Ensures all pods are running and deployment is available

## Testing

The fix has been thoroughly tested with the following results:

✅ **Image Update**: Image reference updates correctly with new tags  
✅ **Pod Recreation**: New pods are created during deployment  
✅ **Deployment Stability**: All pods are running and ready  
✅ **E2E Test Passes**: Comprehensive test verifies the complete workflow  

**Test Output:**

```txt
rayyan@rayyan-seliya:/mnt/c/Users/RAYYAN/Desktop/kubeflex$ ./test/e2e/test-controller-image-update.sh
+ set -e
+++ dirname ./test/e2e/test-controller-image-update.sh
++ cd ./test/e2e
[...]

PR #590, opened 2025-12-15
🐛 Restore stripping of leading "v" from container image tag

MikeSpreitzer:
## Summary
This PR restores the stripping of a leading "v" from the image tag in the Makefile target `chart`.

This is needed because .github/workflows/goreleaser.yml says

```shell
        make chart IMG=${{ env.REGISTRY }}/${{ env.OPERATOR_IMAGE }}:${{github.ref_name}}
```

and the tag has a leading "v" but the controller-manager container image tag does not.

This stripping was erroneously removed in #519 .

## Related issue(s)

This takes a step towards resolving #588; that Issue will not be fully resolved until a new release is made.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -191,7 +191,7 @@ undeploy: ## Undeploy manager from the K8s cluster specified in ~/.kube/config. 
 
 .PHONY: chart
 chart: manifests kustomize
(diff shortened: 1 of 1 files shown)


pdettori (approved):
/lgtm

Issue #589, opened 2025-12-15
bug: no CI tests the latest release


MikeSpreitzer:
### Describe the bug

Bug #588 was not detected by KubeFlex CI.

### Steps To Reproduce

See above

### Expected Behavior

Some things only happen in the release process. They should be tested by CI.

Issue #541, opened 2025-09-11
feature: Enhance the Helm chart by ensuring that the `kubeflex-system` exists


francostellari:
### Feature Description

Enhance the Helm chart by ensuring that the `kubeflex-system` exists.

This should reduce user errors and simplify/shorten the Helm commands.

### Proposed Solution

Add the creation of the namespace in the Helm template, if it does not already exists

### Want to contribute?

- [ ] I would like to work on this issue.

MikeSpreitzer:
This seems to not be working now.

```console
mspreitz@mjs13 kubeflex % helm upgrade --install kubeflex-operator oci://ghcr.io/kubestellar/kubeflex/chart/kubeflex-operator --version v0.9.1  
Release "kubeflex-operator" does not exist. Installing it now.
Pulled: ghcr.io/kubestellar/kubeflex/chart/kubeflex-operator:v0.9.1
Digest: sha256:2546bb97b81437ee8de0f14f6a6f322e669eb288e813b4c0ecc3de2b6995dfe8
Error: namespaces "kubeflex-system" not found
namespaces "kubeflex-system" not found
namespaces "kubeflex-system" not found
namespaces "kubeflex-system" not found
namespaces "kubeflex-system" not found
namespaces "kubeflex-system" not found
namespaces "kubeflex-system" not found
namespaces "kubeflex-system" not found
namespaces "kubeflex-system" not found
```

MikeSpreitzer:
@shivansh-source: fixing this Issue is _not_ required in order for #594 to work. See how the new scripting in #594 ensures that the `kubeflex-system` namespace exists before instantiating the Helm chart?

shivansh-source:
the previous implementation relied on Helm lookup to conditionally create the `kubeflex-system` namespace. This breaks in OCI installs and in clusters where Helm lacks permission to read namespaces, causing installs to fail with` namespaces "kubeflex-system" not found.`

MikeSpreitzer:
Oh. What I am demonstrating is that #542 is not in release v0.9.1. We knew that.

Never mind.

PR #594, opened 2025-12-17
✨  Generalize E2E to support testing released Kubeflex artifacts

shivansh-source:
## Summary
This PR generalizes the E2E test setup so that the same E2E suite can be run against released Kubeflex artifacts, not only the local source tree.

Previously, E2E always built and installed Kubeflex from local sources. As a result, issues that existed only in released Helm charts or images could pass CI/E2E but fail for users installing a release.

This change makes the installation mode selectable while keeping default PR behavior unchanged.

1. Generalized E2E install path

The E2E setup now supports two modes:

Local source (default)
Builds binaries and images locally
Installs the locally generated chart
Used for PRs and normal CI runs
Released artifacts (opt-in)
Installs the released kubeflex-operator Helm chart from GHCR (OCI)

Uses an explicit release version

Mirrors how KubeFlex is consumed by the core KubeStellar chart
Release (opt-in):
install Kubeflex using the published Helm chart and images
supports a literal release (e.g. vX.Y.Z) or the alias latest. 

2. Explicit command-line control (no env vars)

Release testing is selected via a command-line argument (--release), not environment variables.
This keeps the behavior explicit at the call site and avoids hidden dynamic scoping.

3. Generalized existing E2E workflow

The existing test-e2e workflow is extended with an optional workflow_dispatch input:
No input → unchanged local-source E2E
release=<version> → E2E against released artifacts

##  how to run 
Local source (unchanged):

`test/e2e/run.sh`

Released artifacts:

`test/e2e/run.sh --release v0.9.1` for example
or 
`test/e2e/run.sh --release vX.Y.Z`
Screenshot 2025-12-19 185659
Screenshot 2025-12-19 185816
Screenshot 2025-12-19 185853
Screenshot 2025-12-19 185916
 this is after i ran `test/e2e/run.sh --release v0.9.1`

Fixes #
#589
#588

### DIFF
--- a/test/e2e/README.md
+++ b/test/e2e/README.md
@@ -6,7 +6,7 @@ End-to-end tests that can be used manually by a contributor or be triggered by C
 `kind`,`jq` and  `kubectl` are required to run the test.
 
 ## Run E2E tests manually
-
+ #### The E2E test suite can be run in the following modes:
   - local source builds 
   - a released KubeFlex version, specified with the `--release` option
    
--- a/test/e2e/setup-kubeflex.sh
+++ b/test/e2e/setup-kubeflex.sh
@@ -84,10 +84,6 @@ else
     fi
 
     echo "Installing kubeflex release ${release}"
-       # Ensure release tag has leading 'v'
-     if [[ "${release}" != v* ]]; then
-       release="v${release}"
-     fi
     helm install kubeflex \
       oci://ghcr.io/kubestellar/kubeflex/chart/kubeflex-operator \
       --version "${release}"


test/e2e/setup-kubeflex.sh
```diff
@@ -46,7 +48,19 @@ make ko-local-build
...
+else
+  echo "Installing kubeflex release: ${KUBEFLEX_RELEASE}"
+
+  helm repo add kubeflex https://kubestellar.github.io/kubeflex
+  helm repo update
 
+  helm install kubeflex kubeflex/kubeflex \
+    --namespace kubeflex-system \
+    --create-namespace \
+    --version "${KUBEFLEX_RELEASE}"
```
MikeSpreitzer: It would be better if this aligned more closely with how ks/ks uses the KubeFlex chart. E.g., see https://github.com/kubestellar/kubestellar/blob/9062484ec872310e351bb578a7a257229a979239/core-chart/Chart.yaml#L31-L34 and https://github.com/kubestellar/kubestellar/blob/9062484ec872310e351bb578a7a257229a979239/docs/content/direct/get-started.md#use-core-helm-chart-to-initialize-kubeflex-and-create-its-and-wds . Note that these are instantiating the charts into the default namespace.

.github/workflows/test-e2e-released.yml
```diff
@@ -0,0 +1,30 @@
+name: E2E (Released Kubeflex)
```
MikeSpreitzer: It is best if there is a very close correspondence between the workflow name and the name of the file containing the workflow. For example, the proposed filename would work well if the workflow name were "Test E2E Released".

.github/workflows/test-e2e-released.yml
MikeSpreitzer: Do we really need a new workflow? Why not generalize the existing E2E testing workflow?

shivansh-source: actually in my last pr u specifically asked to introduce a workflow  that tests a given release (accepting the alias "latest" or a literal release. so that why i made a new workflow

shivansh-source:
@MikeSpreitzer i did all the changes and gave proof of sucessful testing .

test/e2e/setup-kubeflex.sh
```diff
@@ -46,7 +60,15 @@ make ko-local-build
...
 make install-local-chart
-
+:
+:
+else
+  echo "Installing kubeflex release: ${release}"
+  kubectl create namespace kubeflex-system --dry-run=client -o yaml | kubectl apply -f -
+  helm install kubeflex \
+   oci://ghcr.io/kubestellar/kubeflex/chart/kubeflex-operator \
+    --version "${release}"
```
MikeSpreitzer: This script needs to interpret "latest".

```console
mspreitz@mjs13 kubeflex % helm install kubeflex \
   oci://ghcr.io/kubestellar/kubeflex/chart/kubeflex-operator \
    --version latest
Error: INSTALLATION FAILED: unable to locate any tags in provided repository: oci://ghcr.io/kubestellar/kubeflex/chart/kubeflex-operator
```

test/e2e/run.sh
MikeSpreitzer: The documentation about this file (README.md) also needs to change.

MikeSpreitzer:
Where is the evidence of successful testing of the new functionality?

shivansh-source:
@MikeSpreitzer The kubeflex Helm chart is published in a way that does not support tag listing from GHCR, so Helm cannot infer latest. To support latest, the script resolves it to the most recent GitHub Release tag and installs that explicit version.

shivansh-source:
the updated e2e logic detected the bug  that previous ci have missed . it detected the release only regression that was previoiusly missed . I hope v0.9.3 doesnt have this bug .

test/e2e/setup-kubeflex.sh
```diff
@@ -46,7 +60,33 @@ make ko-local-build
...
+:
+else
+    echo "Installing kubeflex release: ${release}"
+    kubectl create namespace kubeflex-system --dry-run=client -o yaml | kubectl apply -f -
+
+    if [[ "${release}" == "latest" ]]; then
+        echo "Resolving latest kubeflex release from GitHub"
+        release="$(curl -fsSL https://api.github.com/repos/kubestellar/kubeflex/releases/latest \
+          | grep '"tag_name"' \
+          | cut -d '"' -f 4)"
```
MikeSpreitzer: Wouldn't `jq .tag_name` be better?

test/e2e/setup-kubeflex.sh
```diff
@@ -15,7 +15,23 @@
...
 set -e # exit on error
-
+release=""
+while [[ $# -gt 0 ]]; do
+  case "$1" in
+    --release)
+    if [[ $# -lt 2 ]]; then
+    echo "Error: --release requires a value (e.g. v0.9.2 or latest)"
+    exit 1
+    fi
```
MikeSpreitzer: The indenting of the new code here is wrong in two ways. The "if" and "fi" should be as indented as the command following the "fi". The lines between "if" and "fi" should be further indented.

image

test/e2e/README.md
```diff
@@ -1,15 +1,19 @@
...
+  - local source builds 
+  - a released KubeFlex version, specified with the `--release` option
+   
+When a release is specified, the tests install KubeFlex from the published Helm chart. The special value latest installs the most recent released version, while a literal version installs that specific release.
 
-Cleaning up:
-```
-test/e2e/cleanup.sh
+From the root directory of this git repository, you can run:
+```shell
```
MikeSpreitzer: Put a blank line before here. Some Markdown linters require it. The reason being: this starts a new paragraph, and that should be clear.

shivansh-source:
@MikeSpreitzer should i start working on #541 maybe see if i can do some thing so this pr can be fixed . ?

MikeSpreitzer:
We will need evidence that the revised PR works.

test/e2e/setup-kubeflex.sh
```diff
@@ -46,7 +64,34 @@ make ko-local-build
...
+        fi
+
+        echo "Resolved latest release to ${release}"
+    fi
 
+    echo "Installing kubeflex release ${release}"
+       # Ensure release tag has leading 'v'
+     if [[ "${release}" != v* ]]; then
+       release="v${release}"
+     fi
```
MikeSpreitzer: In my opinion this is unneeded complexity. I think that it is reasonable to expect the user to provide a valid release string.

katara-Jayprakash:
@shivansh-source, we're currently focused on the Kubeflex 1.32 release update. Could you resolve this by tomorrow?

shivansh-source:
[for0.9.1.txt](https://github.com/user-attachments/files/24415691/for0.9.1.txt)
full logs for 0.9.1

MikeSpreitzer (commented):
More work remains to be done but this is good enough to merge.
/lgtm
/approve
Thanks!

PR #612, opened 2026-01-09
🐛 Warn users about broken v0.9.2 release

### LINKED ISSUE #607: bug: Remove non working release v0.9.2
### Describe the bug

Since release v0.9.2 is not working, and especially because it is the latest one that would be picked up by a user, it should be removed from the list of available releases or at the very list a new bold comment added to it's release notes that it is broken

### Steps To Reproduce

It's a know issue that the container images referenced by the chart has a mismatch in the use of the prepending "v"

### Expected Behavior

v0.9.2. should not be so prominently visible to a unknowing user

### Additional Context

_No response_

xonas1101:
## Summary
This PR does not block or prevent users from installing v0.9.2.
Instead, it adds clear warnings and safeguards to protect users from accidentally running into a known broken release.

In users.md, a dedicated warning section now highlights that v0.9.2 is broken, so users are informed before following outdated instructions.

In test/e2e/setup-kubeflex.sh, we only log warning messages when:

--release v0.9.2 is used

or when 

--release latest resolves to v0.9.2
In both cases, the script warns the user and falls back to v0.9.1 for safety.
## Related issue(s)

Fixes #607


xonas1101:
Hi @MikeSpreitzer, I have performed the requested changes. Also, I somehow got rid of that "ghost" diff.

PR #625, opened 2026-01-12
🐛 Condition update test on being local

### LINKED ISSUE #624: bug: E2E test of a release bogusly tests update with local chart
### Describe the bug

`test/e2e/run.sh --release v0.9.1` fails because it invokes `helm update` a second time, to check that `helm upgrade` does indeed work, but this always uses the local chart --- which is wrong for testing a release.

### Steps To Reproduce

Shown above

### Expected Behavior

Testing of a release works.

### Additional Context

_No response_

MikeSpreitzer:
## Summary
This PR updates the E2E test suite to exclude the testing of controller image update when the subject under test is not the local copy, because that test does not make sense in that case.

## Related issue(s)

Fixes #624

### DIFF
--- a/test/e2e/run.sh
+++ b/test/e2e/run.sh
@@ -60,7 +60,12 @@ else
 fi
 
 ${SRC_DIR}/manage-type-k8s.sh
-${SRC_DIR}/test-controller-image-update.sh
+
+if [ -z "${release}" ]; then
+    # This test is only appropriate when testing the local copy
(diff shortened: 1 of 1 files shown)


MikeSpreitzer:
Successful run on local copy (no release): https://github.com/kubestellar/kubeflex/actions/runs/20927996155/job/60131122371?pr=625

PR #670, opened 2026-04-03
Fix: bug: bug: False positive readiness check when API server watch channels close unexpectedly

### LINKED ISSUE #669: bug: bug: False positive readiness check when API server watch channels close unexpectedly
### Describe the bug

@clubanderson @MikeSpreitzer  In `pkg/util/status_check.go`, the functions `WaitForDeploymentReady`, `WaitForStatefulSetReady`, and `WaitForNamespaceDeletion` establish a watch on the Kubernetes API. However, if the watch channel (`watcher.ResultChan()`) is closed by the API server (e.g., due to a timeout, abruptly dropped connection, or API server restart) before the resource is actively ready, the functions gracefully catch the closed channel but incorrectly return `nil`.

Because `nil` implies success, the calling controller inherently assumes the Deployment/StatefulSet/Namespace is fully ready or successfully deleted, even if it is not. This can lead to downstream race conditions where the reconciler proceeds with post-creation hooks or subsequent configuration logic while the required infrastructure is still in an incomplete state.

### Steps To Reproduce

1. Trigger a control plane reconciliation that relies on `WaitForDeploymentReady` or `WaitForStatefulSetReady`.
2. While the function is blocking on `<-watcher.ResultChan()`, simulate an API server connection drop or wait for the watch timeout to be reached before the replica counts match.
3. Observe that the function returns `nil` (success).
4. See the reconciler proceed with subsequent steps despite the resource not actually being ready.

### Expected Behavior

If the watch channel closes unexpectedly before the explicit readiness conditions are met (e.g. `deploy.Status.ReadyReplicas == deploy.Status.Replicas`), the function should return an explicit error (like `fmt.Errorf("watch channel closed before resource became ready")`) rather than `nil`. This allows the caller to catch the error and cleanly retry or re-establish the watch.

### Additional Context

Code snippet demonstrating the flaw in `pkg/util/status_check.go`:
```go
for {
	event, ok := <-watcher.ResultChan()
	if !ok {
		// BUG: Returns nil (success) if the watch is finalized/closed!
		return nil 
	}
// ...

### PR DESCRIPTION
**Pull request by gabrnavarro**
## What changed

In `pkg/util/status_check.go`, the three watch-based functions now return an explicit error instead of `nil` when the watch channel closes before the expected condition is met:

- `WaitForDeploymentReady`: returns `fmt.Errorf("watch channel closed before deployment %s/%s became ready", ...)`
- `WaitForStatefulSetReady`: returns `fmt.Errorf("watch channel closed before statefulset %s/%s became ready", ...)`
- `WaitForNamespaceDeletion`: returns `fmt.Errorf("watch channel closed before namespace %s was deleted", ...)`

The parameter types were also changed from `kubernetes.Clientset` (concrete struct) to `kubernetes.Interface` to allow injection of a fake client in tests.

A new test file `pkg/util/status_check_test.go` covers both the bug scenario (channel closed early → error) and the happy path (condition met → nil) for all three functions.

## Why

When the Kubernetes API server closes a watch channel prematurely (timeout, dropped connection, restart), the original code returned `nil`, signaling success to callers. This caused the reconciler to proceed as if the Deployment/StatefulSet was ready or the Namespace was deleted, leading to race conditions and incorrect downstream behavior.

Fixes #669

## Testing

- `TestWaitForDeploymentReady_ChannelClosedBeforeReady`: closes watch channel before deployment is ready, asserts error returned
- `TestWaitForDeploymentReady_ReadyBeforeChannelClose`: sends a ready deployment event, asserts nil returned
- `TestWaitForStatefulSetReady_ChannelClosedBeforeReady`: same pattern for StatefulSet
- `TestWaitForNamespaceDeletion_ChannelClosedBeforeDeletion`: closes watch channel before namespace deleted, asserts error
- `TestWaitForNamespaceDeletion_DeletedSuccessfully`: sends Terminating then Deleted events, asserts nil

All existing tests continue to pass (`go test ./...`).

### DIFF
--- a/cmd/kflex/delete/delete.go
+++ b/cmd/kflex/delete/delete.go
@@ -92,7 +92,7 @@ func ExecuteDelete(cp common.CP, chattyStatus bool) error {
 		return fmt.Errorf("error getting kf client: %v", err)
 	}
 	util.PrintStatus(fmt.Sprintf("Waiting for control plane %s to be deleted...", cp.Name), done, &wg, chattyStatus)
-	util.WaitForNamespaceDeletion(*clientsetKflex, util.GenerateNamespaceFromControlPlaneName(cp.Name))
+	util.WaitForNamespaceDeletion(clientsetKflex, util.GenerateNamespaceFromControlPlaneName(cp.Name))
 
 	if controlPlane.Spec.Type != tenancyv1alpha1.ControlPlaneTypeHost &&
 		controlPlane.Spec.Type != tenancyv1alpha1.ControlPlaneTypeExternal {
--- a/pkg/reconcilers/shared/postcreate_hook.go
+++ b/pkg/reconcilers/shared/postcreate_hook.go
@@ -32,9 +32,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	clog "sigs.k8s.io/controller-runtime/pkg/log"
 
+	"errors"
 	"github.com/kubestellar/kubeflex/api/v1alpha1"
 	"github.com/kubestellar/kubeflex/pkg/util"
-	"errors"
 )
 
 const (
--- a/pkg/util/status_check_test.go
+++ b/pkg/util/status_check_test.go
@@ -1,0 +1,143 @@
+/*
+Copyright 2023 The KubeStellar Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import (
+	"testing"
+
+	v1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/kubernetes/fake"
+	k8stesting "k8s.io/client-go/testing"
+)
+
+func int32Ptr(i int32) *int32 { return &i }
+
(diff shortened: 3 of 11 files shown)

### CODE REVIEW

File: pkg/reconcilers/shared/postcreate_hook.go [hunk 1]
```diff
@@ -32,9 +32,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	clog "sigs.k8s.io/controller-runtime/pkg/log"
 
+	"errors"
```
MikeSpreitzer, line 35: This should be listed in the first group of imports, those from the standard library.

cmd/kflex/delete/delete.go
```diff
@@ -92,7 +92,7 @@ func ExecuteDelete(cp common.CP, chattyStatus bool) error {
 		return fmt.Errorf("error getting kf client: %v", err)
 	}
 	util.PrintStatus(fmt.Sprintf("Waiting for control plane %s to be deleted...", cp.Name), done, &wg, chattyStatus)
-	util.WaitForNamespaceDeletion(*clientsetKflex, util.GenerateNamespaceFromControlPlaneName(cp.Name))
+	util.WaitForNamespaceDeletion(clientsetKflex, util.GenerateNamespaceFromControlPlaneName(cp.Name))
```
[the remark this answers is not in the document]
MikeSpreitzer: I think that this is a valid complaint. And adjacent to it is another bug: because we have no quantitative guarantees of timing, the object in question might already be gone before the watch even starts --- in which case the watch will never see the object's deletion. A correct way to both recover from a terminated watch and have no gaps while there is no watch is to use an _informer_ rather than direct coding of watches. Another would be polling.

pkg/util/status_check_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+	v1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/kubernetes/fake"
+	k8stesting "k8s.io/client-go/testing"
+)
+
+func int32Ptr(i int32) *int32 { return &i }
```
MikeSpreitzer, line 30: FYI, there is already a utility for this, in k8s.io/utils/ptr .

MikeSpreitzer:
I think that this is a step forward, although an incomplete one (as mentioned above).

/lgtm
/approve
Thanks!

